PR #18, opened 2021-07-14
Initial skeleton with recursive filepath parsing.

evetion:


hydrolib/core/basemodel.py
```diff
@@ -0,0 +1,127 @@
...
+    @classmethod
+    def validate(cls: Type["FileModel"], value: Any) -> "FileModel":
+        # Enable initialization with a Path.
+        if isinstance(value, Path):
+            # Pydantic Model init requires a dict
+            value = {"filepath": value}
+        return super().validate(value)
+
+    def _load(self, filepath: Path):
+        if Path(filepath).is_file():
```
priscavdsluis, line 59:
```suggestion
        if filepath.is_file():
```

Can't we do the above?

hydrolib/core/basemodel.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def save(self, folder: Path, force=False):
+        """Save model and child models to their set filepaths.
+
+        For models with an unset filepath, we generate one based
+        on the given `folder`.
+
+        If `force` is set, we override the folder part of
+        already set filepaths, which can be used to copy complete models.
+        """
```
BeardedPlatypus, line 73:
```suggestion
        """Save model and child models to their set filepaths.

        For models with an unset filepath, we generate one based
        on the given `folder`.

        If `force` is set, we override the folder part of
        already set filepaths, which can be used to copy complete models.
        
        Args:
            folder (Path): path to the folder where this FileModel will be stored
            force (bool, optional): Force override the folder part of already set filepath
        """
```

Do we need to document the arguments? or are they too trivial and do we ignore them?

hydrolib/core/basemodel.py
```diff
@@ -0,0 +1,127 @@
...
+        self._get_serializer()(self.filepath, data)
+
+    @classmethod
+    def _parse(cls, path: Path):
+        return cls._get_parser()(path)
+
+    @classmethod
+    def _generate_name(cls):
+        name, ext = cls._filename(), cls._ext()
+        return Path(f"{name}{ext}")
```
BeardedPlatypus, line 111: Should we specify this returns a Path, or should this return str?

hydrolib/core/basemodel.py
```diff
@@ -0,0 +1,127 @@
...
+            self.filepath = folder / self.filepath.name
+
+        # Convert child FileModels first
+        exclude = {"filepath"}
+        filemodel_fields = {}
+        for name, value in self:
+            if isinstance(value, FileModel):
+                filepath = value.save(folder, force)
+                filemodel_fields[name] = filepath
+                exclude.add(name)
```
priscavdsluis, line 88: Does this prevent that the reference to this child file will not be written to the parent file? For example, in the mdu file we reference the network file.

BeardedPlatypus (changes requested):

tests/test_model.py
```diff
@@ -0,0 +1,28 @@
+from pathlib import Path
+
+from hydrolib.core.basemodel import FileModel
+from hydrolib.core.models import DIMR, FMModel
+
+from .utils import test_data_dir, test_output_dir
+
+
+def test_filemodel_load():
+    d = DIMR(filepath="a", model=Path(test_data_dir / "test.mdu"))
```
priscavdsluis, line 10: Did you intend to test with a file that does not exist?

priscavdsluis (changes requested):

hydrolib/core/basemodel.py
evetion: It prevents the child file being returned as {fielda: value, etc} (automatic due to Pydantic), now it is returned as Path("location/of/child") by ourselves.

tests/test_model.py
evetion: Yes, I will make this explicit in the test description.

BeardedPlatypus (approved):

priscavdsluis (approved):

PR #21, opened 2021-07-20
Parse dimr xml files

priscavdsluis:


hydrolib/core/models.py
```diff
@@ -150,3 +153,11 @@ def _get_serializer(cls) -> Callable:
     @classmethod
     def _get_parser(cls) -> Callable:
         return DIMRParser.parse
+
+    @validator("component", pre=True)
+    def validate_component(cls, v):
+        return to_list(v)
+
+    @validator("coupler", pre=True)
```
evetion, line 161: You can combine this one into the previous validator by giving it multiple fields.

hydrolib/core/utils.py
```diff
@@ -13,3 +16,18 @@ def example(a: float, b: float = 1.0) -> float:
...
+    """Puts the specified item in a list if it is an instance of `dict`.
+
+    Attributes:
+        item: The item to put in a list.
+
+    Returns:
+        List: If the item is a dictionary, a list with the specified item; otherwise, the item.
+    """
+
+    if isinstance(item, dict):
```
evetion, line 31: Nitpick: This doesn't always return a `list`. You fix it by `if not isinstance(item, list)`.

PR #23, opened 2021-07-22
Add parser for INI files

priscavdsluis:


evetion:
A few thoughts @BeardedPlatypus @priscavdsluis 

### Comments
With regards to storing comments, I feel we should represent the ini structure as closely as possible, so setting 
```ini
[group]
a=1
```

should be something like `INI(group={"a":1})` on init and `ini.group.a = 1` after. This doesn't leave room for comments however.

I've tried my hand at storing the comments inside a standard type, so creating `CommentInt`, `CommentFloat` etc. While this does work it's a lot of boilerplate and probably confusing for the users. I think storing the comments at the Model itself is better, with attributes that shadow the actual data fields. This is something that we can add afterwards.

### Configparser
INI parsing is normally done by [`ConfigParser`](https://github.com/python/cpython/blob/main/Lib/configparser.py), a Python built in. It doesn't support repeated sections though, but one could theoretically override this by setting your own dict type that overrides `__setitem__`, but this is not that straightforward. It would however save a lot of boilerplate.

```python
configparser.RawConfigParser(
        strict=False, dict_type=OurOwnMultiDict, allow_no_value=True
    )
```

BeardedPlatypus:
@evetion 
I personally rather not inherit from Configparser, as I feel like the deltares flavour of ini files is significantly different that it trying to override it to behave the way we want will not make the code very readable. 

The current approach I would like to take is separate the parsing of deltares ini-ish files in a scanner, which would create a simple representation of keyvalue pairs, comments, sections etc, with their corresponding line numbers. Then transforming them through pydantic validators in the actual representations which we expose.

hydrolib/core/io/structure/models.py
```diff
@@ -0,0 +1,129 @@
...
+            alias="xCoordinates",
+        )
+        y_coordinates: Optional[str] = Field(
+            "y-coordinates of the location of the structure. (number of values = numCoordinates)",
+            alias="yCoordinates",
+        )
+
+    comments: Comments = Field(default_factory=Comments)
+
+    header: Literal["Structure"] = "Structure"
```
evetion, line 34: This probably shouldn't be public?

hydrolib/core/io/structure/models.py
```diff
@@ -0,0 +1,129 @@
...
+
+class FlowDirection(str, Enum):
+    none = "none"
+    positive = "positive"
+    negative = "negative"
+    both = "both"
+
+
+class Weir(Structure):
+    class Comments(Structure.Comments):
```
evetion, line 56: That's some odd inheritance, but I guess it otherwise wouldn't actually inherit your abc one?

evetion: Might we also move the class comments to underneath the normal fields? Ah, that would break the comments field itself I guess.

hydrolib/core/io/structure/models.py
```diff
@@ -0,0 +1,129 @@
...
+    allowed_flow_direction: FlowDirection = Field(alias="allowedFlowdir")
+
+    number_of_levels: int = Field(alias="numLevels")
+    y_values: List[float] = Field(alias="yValues")
+    z_values: List[float] = Field(alias="zValues")
+    crest_level: float = Field(alias="crestLevel")
+    discharge_coefficient: float = Field(alias="dischargeCoeff")
+
+    @validator("y_values", "z_values", pre=True)
+    def split_value_lists(cls, v):
```
evetion, line 126: You wanted to re-use right? Check https://pydantic-docs.helpmanual.io/usage/validators/#reuse-validators. Or if you think that's ugly, at least split it to a utils and call it from here.

evetion (commented):
Can you also add a test which initializes a Weird from scratch (without parsing) and also sets the comments for a field or two? Good to have that public API in here to understand what you've made.

hydrolib/core/io/structure/models.py
```diff
@@ -0,0 +1,129 @@
...
+            alias="zValues",
+        )
+        crest_level: Optional[str] = Field(
+            "Crest level of weir (m AD).", alias="crestLevel"
+        )
+        discharge_coefficient: Optional[str] = Field(
+            "Discharge coefficient c_e (-).", alias="dischargeCoeff"
+        )
+
+    comments: Comments = Field(default_factory=Comments)
```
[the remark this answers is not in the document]
BeardedPlatypus: Ah right, I think I might even be able to just call 

```python
comments: Comments = Comments()
```

I will fix that.

hydrolib/core/io/structure/models.py
BeardedPlatypus: Yeah unfortunately, the `Comments` need to be defined before the field. 
I wanted to make sure that the `Weir.Comments` still has the fields of `Structure.Comments`. I do agree it looks a bit odd, but I'm not sure if there is a better way to do this. We could add the additional fields through some class method which would be called in the IniBasedModel, and then the fields would be added at that point. However, if my understanding of pylance and other ide intellisense is correct, it would be hard to automatically retrieve the fields (and mypy / pylance would probably complain). 

I personally would say that the additional type hinting you get by this inheritance trumps how it looks, seen as much as the help with the fields by end users. But I might be wrong there.

hydrolib/core/io/structure/models.py
BeardedPlatypus: Probably a mistake in my configuration of the `IniBasedModel`, but when I make this private, pydantic does not parse the classes correctly any more.

evetion (approved):

PR #31, opened 2021-07-26
Add network functionality

evetion:


grongen:
The code contains the core functionality. Some comments and docstring still need to be added. It is Ready for review, but does still contain an issue:
The "Network" class "hydrolib.core.io.net.models.Network" should be merged into the "hydrolib.core.models.Network" class that inherits from BaseModel. This still gives some problems due to pydantic. It is added as a new issue.

hydrolib/core/io/net/models.py
```diff
@@ -0,0 +1,916 @@
...
+            )
+            link_coords = np.stack(
+                [
+                    nodes1d[self._link1d2d.link1d2d[:, 0]],
+                    faces2d[self._link1d2d.link1d2d[:, 1]],
+                ],
+                axis=1,
+            )
+            lc_link1d2d = LineCollection(link_coords, **links1d2d_kwargs)
+            ax.add_collection(lc_link1d2d)
```
BeardedPlatypus, line 916: I would move this to a separate utility which takes a `Network` as an argument, rather than have it as a property of the Network itself. This would allow us to drop the dependency on matplotlib for the core models, and instead localise all plotting related logic to a separate sub-model.

hydrolib/core/io/net/models.py
```diff
@@ -0,0 +1,916 @@
...
+            ax.add_collection(lc_mesh1d)
+
+        # Mesh 2d
+        if not self._mesh2d.empty():
+            nodes2d = np.stack(
+                [self._mesh2d.mesh2d_node_x, self._mesh2d.mesh2d_node_y], axis=1
+            )
+            edge_nodes = self._mesh2d.mesh2d_edge_nodes
+            lc_mesh2d = LineCollection(nodes2d[edge_nodes], **mesh2d_kwargs)
+            ax.add_collection(lc_mesh2d)
```
BeardedPlatypus, line 901: I think you could abstract this logic into a separate private function if we would unify the mesh1d_node_x and mesh2d_node_x to a single node_x (same_for node_y):

```python
class MeshBased(Protocol):
    node_x
    node_y
    edge_nodes

    def empty(self):
        """empty check."""

def add_mesh_to_plot(mesh: MeshBased, ax, styling_kwargs):
    if mesh.empty():
        return

    nodes = np.stack([mesh.node_x, mesh.node_y], axis=1)
    lc_mesh2d = LineCollection(nodes[mesh.edge_nodes], **styling_kwargs)
    ax.add_collection(lc_mesh2d)
```

then the internal code could just be 
```python
def plot(
        self,
        ax: matplotlib.axes._subplots.AxesSubplot,
        mesh1d_kwargs: dict = None,
        mesh2d_kwargs: dict = None,
        links1d2d_kwargs: dict = None,
    ) -> None:

        if mesh1d_kwargs is None:
            mesh1d_kwargs = {"color": "C3", "lw": 1.0}
        if mesh2d_kwargs is None:
            mesh2d_kwargs = {"color": "C0", "lw": 0.5}
        if links1d2d_kwargs is None:
            links1d2d_kwargs = {"color": "k", "lw": 1.0}
        
        add_mesh_to_plot(self._mesh1d, ax, mesh1d_kwargs)
        add_mesh_to_plot(self._mesh2d, ax, mesh2d_kwargs)

        # Links
        if not self._link1d2d.empty():
            faces2d = np.stack(
                [self._mesh2d.mesh2d_face_x, self._mesh2d.mesh2d_face_y], axis=1
            )
            link_coords = np.stack(
                [
                    nodes1d[self._link1d2d.link1d2d[:, 0]],
                    faces2d[self._link1d2d.link1d2d[:, 1]],
                ],
                axis=1,
            )
            lc_link1d2d = LineCollection(link_coords, **links1d2d_kwargs)
            ax.add_collection(lc_link1d2
```

For consistency sake you could move the links to an `add_links_to_plot` so all function operate at the same abstraction.

hydrolib/core/io/net/models.py
```diff
@@ -0,0 +1,916 @@
...
+
+    mesh2d_edge_x: np.ndarray = np.empty(0, dtype=np.double)
+    mesh2d_edge_y: np.ndarray = np.empty(0, dtype=np.double)
+    mesh2d_edge_z: np.ndarray = np.empty(0, dtype=np.double)
+    mesh2d_edge_nodes: np.ndarray = np.empty((0, 2), dtype=np.int32)
+
+    mesh2d_face_x: np.ndarray = np.empty(0, dtype=np.double)
+    mesh2d_face_y: np.ndarray = np.empty(0, dtype=np.double)
+    mesh2d_face_z: np.ndarray = np.empty(0, dtype=np.double)
+    mesh2d_face_nodes: np.ndarray = np.empty((0, 0), dtype=np.int32)
```
BeardedPlatypus, line 60: I think we could drop the `mesh2d_` prefix, this would also make it easier to define protocols that work on both the 1d and 2d meshes.

hydrolib/core/io/net/models.py
```diff
@@ -0,0 +1,916 @@
...
+
+    def empty(self):
+        return self.mesh2d_node_x.size == 0
```
BeardedPlatypus, line 63:
```suggestion
    def empty(self) -> bool:
        return self.mesh2d_node_x.size == 0
```

Nitpicking: I would personally always prefix boolean functions with either `is_` or `has_`, in this case `is_empty`.

hydrolib/core/io/net/models.py
```diff
@@ -0,0 +1,916 @@
...
+
+    def read_file(self, file: Path) -> None:
+
+        # Create reader and read 2d
```
BeardedPlatypus, line 67:
```suggestion
        """Create reader and read 2d
        
        Args:
            file (Path): the path to the file to read.
        """
```

hydrolib/core/io/net/models.py
```diff
@@ -0,0 +1,916 @@
...
+
+        self.meshkernel.mesh2d_set(mesh2d)
+
+    def get_mesh2d(self) -> mk.Mesh2d:
+        """Return mesh2d from meshkernel"""
+        return self.meshkernel.mesh2d_get()
+
+    # def create_triangular(self, polygon: mk.GeometryList, mesh2d_triangular_options: IntEnum) -> None:
+    #     """Creates a 2d mesh within a polygon."""
+    #     pass
```
BeardedPlatypus, line 87: Can we remove this?
```suggestion
```

BeardedPlatypus (changes requested):
I ran black and added meshkernel and netCDF4 as dependency, and matplotlib as a dev dependency. 
Note that we also have 2 bugs which I believe are missing the assert statements, which could be easily fixed.

I'm not done yet, so I will probably add additional review comments, but I wanted to store already what I have atm.

.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
...
+            "type": "python",
+            "request": "launch",
+            // "program": "D:/Documents/GitHub/HYDROLIB-core/debug.py",
+            "program": "D:/Documents/GitHub/HYDROLIB-core/tests/test_net.py",
+            // "cwd": "D:/Documents/GitHub/HYDROLIB-core",
+            "console": "integratedTerminal",
+            "justMyCode": false
+        }
+    ]
+}
```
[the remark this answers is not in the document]
grongen: Agree, add to .gitignore

debug.py
```diff
@@ -0,0 +1,34 @@
...
+links1d2d_kwargs = dict(lw=0.75, color="black", zorder=3)
+
+m.plot(ax, mesh1d_kwargs=mesh1d_kwargs, mesh2d_kwargs=mesh2d_kwargs, links1d2d_kwargs=links1d2d_kwargs)
+
+ax.autoscale()
+ax.set_aspect(1.0)
+
+plt.show()
+
+print()
```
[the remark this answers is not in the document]
grongen: This can be removed altogether

debug2.py
```diff
@@ -0,0 +1,75 @@
...
+
+    fig, ax = plt.subplots()
+    edge_nodes = mesh2d_output.edge_nodes.reshape((-1, 2))
+    edge_coords = np.dstack([mesh2d_output.node_x[edge_nodes], mesh2d_output.node_y[edge_nodes]])
+    lc = LineCollection(edge_coords)
+    ax.add_collection(lc)
+    ax.autoscale_view()
+    plt.show()
+
+    print("klaar")
```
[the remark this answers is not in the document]
grongen: Same here, remove this file from the branch. It is replaced by test_net

hydrolib/core/io/net/models.py
```diff
@@ -0,0 +1,741 @@
...
+from pydantic.types import PyObject
+from pydantic.utils import ValueItems
+
+from hydrolib.core import __version__
+from hydrolib.core.basemodel import BaseModel
+from hydrolib.core.io.net.reader import UgridReader
+from hydrolib.core.io.net.writer import UgridWriter
+
+import matplotlib
+from matplotlib.collections import LineCollection
```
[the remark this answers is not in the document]
grongen: For plotting functionality (which we should provide I think), at least an axes object is needed. This is a function argument, so no problem (just like in MeshKernelPy). A problem arises when plotting large networks. This is very slow when each individual segment is separately plotted, so using a LineCollection is convenient, which required matplotlib.

An alternative is creating a single line, where all individual segments are seperated with [np.nan, np.nan]. This is a bit hacky, but it works.

Something like ``lc = LineCollection(segments)`` and ``ax.add_collection(lc)`` could be substituted by:

```
def create_separated_line(segments: List[np.ndarray]) -> np.ndarray:
    segments_copy = segments[:]
    separator = np.array([[np.nan, np.nan]])
    for i in range(len(segments)-1, 0, -1):
        segments_copy.insert(i, separator)
    return np.concatenate(segments_copy)
```
and ``ax.plot(*create_single_line(segments).T)``

hydrolib/core/io/net/models.py
grongen: Agreed. In time perhaps a seperate plotting for 1d and 2d could be desired, but we can always change it then.

hydrolib/core/io/net/models.py
grongen: It is indeed a bit wordy, but i wouldn't remove the prefix. In the 1D mesh the distinction is made to distinguish between mesh1d and network1d. From this perspective, and the consistency with the Ugrid naming (which is followed as much as possible) I would leave the mesh2d for clarity.

hydrolib/core/io/net/models.py
grongen: See for the alternative to LineCollection the proposed function a few comments above.

hydrolib/core/io/net/models.py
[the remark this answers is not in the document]
BeardedPlatypus: After discussion with @evetion, we have decided to remove theis logic, and instead move it to a tutorial. We can add such logic in the future to the HYDROLIB package.

BeardedPlatypus:
I have added #43 as a follow up to document the net.models.py classes. This is not really necessary for the MVP, but is important for the final product.

hydrolib/core/io/net/models.py
BeardedPlatypus: We will leave it as is for now, and might pick it up if it becomes a requirement in the future.

hydrolib/core/io/net/models.py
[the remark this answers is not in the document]
BeardedPlatypus: Moved to a tutorial. We should see whether we want to add this back in the future, or provide in the HYDROLIB package.

BeardedPlatypus (approved):
Several issues still need to be addressed in follow up issues, and as with any code, I think it will be good to stay critical as users start using this library. But for now I think we can merge it back.

PR #36, opened 2021-07-28
Add generic INI serializer

BeardedPlatypus:


hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,246 @@
...
+        )
+        return chain(header_iterable, sections_iterable)
+
+
+def write_ini(
+    path: Path, document: Document, config: Optional[SerializerConfig] = None
+) -> None:
+    """Write the provided document to the specified path
+
+    If the provided path already exists, it will be overwritten. If the parent fodels
```
priscavdsluis, line 225:
```suggestion
    If the provided path already exists, it will be overwritten. If the parent folders
```

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,246 @@
...
+            Serializer._interweave(
+                (self._serialize_section(section) for section in document.sections),
+                [""],
+            )
+        )
+        return chain(header_iterable, sections_iterable)
+
+
+def write_ini(
+    path: Path, document: Document, config: Optional[SerializerConfig] = None
```
priscavdsluis, line 221:
```suggestion
    path: Path, document: Document, config: Optional[SerializerConfig] = SerializerConfig()
```

hydrolib/core/io/ini/serializer.py
priscavdsluis: Or is it because you are checking for None anyway?

hydrolib/core/io/ini/serializer.py
priscavdsluis: Or is it because you are checking for None anyway?

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,246 @@
...
+
+    serializer = Serializer(config)
+
+    path.parent.mkdir(parents=True, exist_ok=True)
+
+    with path.open("w") as f:
+
+        for line in serializer.serialize(document):
+            f.write(line)
+            f.write("\n")
```
priscavdsluis, line 246:
```suggestion
            f.write(line + "\n")
```

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,246 @@
...
+            property_indent). Defaults to 4.
+        datablock_indent (int):
+            The number of spaces with which datablock rows are indented relative to
+            the section header (i.e. the full indent equals the section_indent plus
+            datablock_indent). Defaults to 8.
+        datablock_spacing (int):
+            The number of spaces between datablock columns. Note that there might be
+            additional offset to ensure . is lined out. Defaults to 4.
+        comment_delimiter (str):
+            The character used to delimit comments.
```
priscavdsluis, line 29:
```suggestion
            The character used to delimit comments. Default to #.
```

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,246 @@
...
+    def total_property_indent(self):
+        return self.section_indent + self.property_indent
+
+    @property
+    def total_datablock_indent(self):
+        return self.section_indent + self.datablock_indent
+
+
+@dataclass
+class _Lengths:
```
priscavdsluis, line 48: So each section has its own value & comment offset? 
Is it desirable that each section has a different offsets?

hydrolib/core/io/ini/models.py
```diff
@@ -64,17 +58,13 @@ class Section(BaseModel):
 
     Attributes:
         header (str): The header (without brackets) of this Section
-        start_line (int): The starting line index of this Section block
-        end_line (int): The end line index of this Section block
         content (List[Union[Property, CommentBlock]]):
             The ordered list of Property and CommentBlock objects
         datablock (Optional[Sequence[Sequence[str]]]):
```
priscavdsluis, line 63: Maybe document what the "dimentions" are? e.g. whether it is rows x colummns or columns x rows.

hydrolib/core/io/ini/serializer.py
BeardedPlatypus: Yup, that's the idea. That was how I understand it, however it would be easy to change in the future if we do need to to change it.

hydrolib/core/io/ini/serializer.py
BeardedPlatypus: It is generally considered bad practice to use mutable objects as default values for calls, because this single default value instance is used among all calls. This would mean that if the call that uses this default value would make changes to the state, each subsequent call would have these changes as well. As such, the general approach is to set the default value to None, and then checking if it is, and creating the default value. 

 In this case it could work to use SerializerConfig directly because the code should not be changing it. However for the sake of consistency I opted to use None here as well.

hydrolib/core/io/ini/models.py
```diff
@@ -185,8 +175,6 @@ def _convert_section(cls, section: Section) -> Dict:
     def _convert_section_to_dict(cls, value: Section) -> Dict:
```
priscavdsluis, line 177: Github won't let me leave a comment there but I notice you have the method _combine_in_dict
and I have something similar with the _node_to_dictionary method in the dimr_parser. Not sure if it's exactly the same though.

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,246 @@
...
+
+    @staticmethod
+    def _calculate_max_length_datablock(
+        datablock: Optional[Sequence[Sequence[str]]],
+    ) -> Optional[Sequence[int]]:
+        if datablock is None or len(datablock) < 1:
+            return None
+
+        def folder(acc: Tuple, row: Sequence[str]) -> Tuple:
+            return tuple(max(curr_max, len(elem)) for curr_max, elem in zip(acc, row))
```
priscavdsluis, line 62: This is a little hard for me to read :p

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,246 @@
...
+        self._config = config
+
+    @staticmethod
+    def _interweave(iterable: Iterable, val: Any) -> Iterable:
+        # Interweave the provided iterable with the provided value:
+        # iterable_element, val, iterable_element, val, ...
+
+        # Note that this will interweave with val without making copies
+        # as such it is the same object being interweaved.
+        return chain.from_iterable(zip(iterable, repeat(val)))
```
priscavdsluis, line 108: so the chain.from_iterable kinda flattens the tuples that come out of the zip?

hydrolib/core/io/ini/serializer.py
priscavdsluis: Thanks for the explanation!

hydrolib/core/io/ini/serializer.py
priscavdsluis: Ok, I always thought the offsets apply to the whole file.

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,246 @@
...
+            self._serialize_comment_block(block) for block in header_blocks
+        )
+        # We separate comment blocks in the header with an empty line.
+        blocks_as_lines = Serializer._interweave(blocks_as_lines, [""])  # type: ignore
+
+        return chain.from_iterable(blocks_as_lines)
+
+    def _serialize_section_header(self, section_header: str) -> Iterable[str]:
+        indent = " " * (self._config.section_indent)
+        return [f"{indent}[{section_header}]"]
```
priscavdsluis, line 129: It's a little nitpicky but I'd personally return the value not as a list. To keep the scope of the function not bother with how it's used later on. But feel free to ignore this, it's a private function anyway.

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,246 @@
...
+            f"{property.value}{value_offset}"
+            if property.value is not None
+            else value_offset
+        )
+
+        comment = f" # {property.comment}" if property.comment is not None else ""
+
+        return [f"{indent}{key}{value}{comment}".rstrip()]
+
+    def _serialize_content(
```
priscavdsluis, line 154: Can there be a comment block between to sections? Is it then "absorbed" by the upper section?

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,246 @@
...
+        return " " * max(max_length - key_length, 0)
+
+    def _serialize_property(
+        self, property: Property, lengths: _Lengths
+    ) -> Iterable[str]:
+        indent = " " * (self._config.total_property_indent)
+        key_offset = Serializer._get_offset(property.key, lengths.max_key_length)
+        key = f"{property.key}{key_offset} = "
+
+        value_offset = Serializer._get_offset(property.value, lengths.max_value_length)
```
priscavdsluis, line 143: Isn't it technically 'value_offset' and 'comment_offset'? respectively

hydrolib/core/io/ini/serializer.py
BeardedPlatypus: Currently this is not possible, every comment block is assumed to be part of the section before it, except for the first comment blocks, which are part of the document header. Unfortunately it is not really possible to make this distinction easily, therefor I made this assumption as it was easiest to implement.

hydrolib/core/io/ini/serializer.py
BeardedPlatypus: The reason I wanted to return an `Iterable` is because it allows me to just use a `chain` call in the calling level. You are right however that a list might not make so much sense. I have replaced the return list with a yield, which works just as well (both here and at a similar return statement in the Serializer).

hydrolib/core/io/ini/serializer.py
BeardedPlatypus: I was thinking it is the offset associated with the key, as it is dependent on the length of the longest key within the section, thus key_offset made sense. Maybe key_whitespace is a better term though. I will change that.

hydrolib/core/io/ini/serializer.py
BeardedPlatypus: it is a fold function, though in the python sense, calling it a reductor might have been a better name. Basically it says

we have our acc, which is short for accumulator, which accumulates the results, i.e. it contains the data at one point in the reduction, and a next row consisting of our sequence of strings, and we return the accumulator at the end of processing this row. If we then put this in the reduce function, with the list of rows and an initial value:

```python
reduce(folder, datablock, (0,) * len(datablock[0]))
```

It will basically for each row execute the new accumulator with the folder, and return those values.

The folder basically calculates the max value for each element in the row and the max length till then for elements in that column. Because both the acc and row have the same elements, say 4 for the sake of the argument, if we zip them together, we get pairs of the previous max size, and the new element. We then calculate for each pair the new max length for that pair by taking the max between the previous max size and the length of the new element. We stuff all these values in a tuple, and we get back our new tuple consisting of the max elements until the new element.

hydrolib/core/io/ini/serializer.py
BeardedPlatypus: yup, that's the idea, and because zip takes elements till one of the iterables is exhausted, we can just repeat val indefinitely and it will stop when iterable is exhausted.

hydrolib/core/io/ini/models.py
evetion: Good to add this note to #40

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,245 @@
...
+    @property
+    def total_property_indent(self) -> int:
+        return self.section_indent + self.property_indent
+
+    @property
+    def total_datablock_indent(self) -> int:
+        return self.section_indent + self.datablock_indent
+
+
+@dataclass
```
evetion, line 47: Everything uses pydantic, and now suddenly you use a dataclass? I'm not against it, it's just weird.

hydrolib/core/io/ini/serializer.py
evetion: If you need this amount of text to explain it, maybe add that to the code. Python code should be readable, so don't be afraid to use more lines. Make sure to explain the why/thoughts behind this design in there something, maybe even in the module docstring.

hydrolib/core/io/ini/serializer.py
```diff
@@ -0,0 +1,245 @@
...
+            return (
+                indent
+                + (" " * self._config.datablock_spacing)
+                .join((serialize_row_element(elem, i)) for elem, i in zip(row, count()))
+                .rstrip()
+            )
+
+        return (serialize_row(row) for row in datablock)
+
+    def _serialize_section(self, section: Section) -> Iterable[str]:
```
evetion, line 189: Not specific to this section, but it would be good to have a call stack diagram for these parsers and serializers. There can be many private methods that are hard to understand because they don't have a context, i.e. where in the process are they used, what assumptions are made etc.

evetion: Maybe something to add to #40 as well.

evetion (approved):

hydrolib/core/io/ini/serializer.py
BeardedPlatypus: I had it originally as a simple data transfer object within the Serializer itself, but it made more sense to move it out when I added the from_section logic. You are right though that for consistency it makes more sense to just use pydantic, unless there is a good reason not too. I will change it.

Issue #26, opened 2021-07-26
Add structures parser


priscavdsluis:
The INI parser/serializer can be used, but separate objects and conversion need to be added.

Issue #5, opened 2021-07-13
Parse model .mdu


evetion:
The INI parser/serializer can be used, but separate objects and conversion need to be added.

PR #39, opened 2021-07-29
Integrated ini parsing into most models.

### LINKED ISSUE #1: Parse boundary conditions .bc

evetion:
Add support for the following `FileModel`s
- FM model def (`FMModel`) on .mdu files
- Boundary conditions (`ForcingModel`) on .bc files
- External forcing (`ExtModel`) on .ext files
- Structures (`StructureModel`) on .ini files
- Roughness (`FrictionModel`) on .ini files
- CrossDef (`CrossDefModel`) on .ini files
- CrossLoc (`CrossLocModel`) on .ini files

- [x] Still have to fix the killed ini parser tests on Weird etc, sorry @BeardedPlatypus 

Closes #1 #5 #25 #26


hydrolib/core/basemodel.py
```diff
@@ -22,7 +31,35 @@ class Config:
...
-        alias_generator = to_lowercase
+        alias_generator = to_key
+
+    def is_file_link(self) -> bool:
+        """Generic attribute for models backed by a file."""
+        return False
+
+    def is_intermediate_link(self) -> bool:
+        """Generic attribute for models that have children fields that could contain files."""
+        return self.is_file_link()
```
BeardedPlatypus, line 42: Could this be determined dynamically, or is that overkill?

hydrolib/core/basemodel.py
```diff
@@ -38,22 +75,45 @@ class FileModel(BaseModel, ABC):
...
+            # If a context is set, use it
+            folder = context_dir.get(None)
+            if folder and not filepath.is_absolute():
+                logger.info(f"Used context to get {folder} for {filepath}")
+                filepath = folder / filepath
+            # Otherwise we're the root filepath
+            # and should set the context
+            else:
+                logger.info(f"Set context to {filepath.parent}")
+                token = context_dir.set(filepath.parent)
```
BeardedPlatypus, line 97: I'm not entirely sure about this. It would make sense for a model to refer to the context of its parent for loading the file, but I vaguely recall something that relative paths of FMModels are always relative to the .mdu, even if they parent is located somewhere else.

Might be good to verify this with @arthurvd .

hydrolib/core/basemodel.py
```diff
@@ -38,22 +75,45 @@ class FileModel(BaseModel, ABC):
...
     filepath: Optional[Path] = None
 
-    def __init__(self, filepath: Optional[Path] = None, *args, **kwargs) -> None:
+    def __init__(self, filepath: Optional[Path] = None, *args, **kwargs):
         """Initialize a model.
 
         The model is empty (with defaults) if no `filepath` is given,
         otherwise the file at `filepath` will be parsed."""
         # Parse the file if path is given
+        token = None
```
BeardedPlatypus, line 84: can we rename this to `context_dir_reset_token`, or something similar, token feels quite vague to the untrained eye.

hydrolib/core/basemodel.py
```diff
@@ -38,22 +75,45 @@ class FileModel(BaseModel, ABC):
...
+            data = self._load(filepath)
             data["filepath"] = filepath
             kwargs.update(data)
         super().__init__(*args, **kwargs)
 
+        # Once the model has been completely initialized
+        # reset the context
+        if token:
+            logger.info(f"Reset context.")
+            context_dir.reset(token)
```
BeardedPlatypus, line 108: If I see a comment line above a bit of code, I'd be inclined to replace those with a separate private function which incorporates the comment. This way your code is self documenting, and it is harder for the code to go out of sync with it's description, e.g.:

```suggestion
        self._reset_context_dir(token)
```
and then add
```python
    def _reset_context_dir(token)
        if token:
            logger.info(f"Reset context.")
            context_dir.reset(token)
```

hydrolib/core/io/bc/models.py
```diff
@@ -136,50 +50,107 @@ def _supports_comments(cls):
...
+                if (
+                    c.__fields__.get("function").default
+                    == v.get("function", "").lower()
+                ):
+                    v = c(**v)
+                    break
+        return v
+
+
+class TimeSeries(ForcingBase):
```
BeardedPlatypus, line 77: I personally feel like a Forcing component has a TimeSeries function, rather than being a TimeSeries.

hydrolib/core/io/bc/models.py
```diff
@@ -136,50 +50,107 @@ def _supports_comments(cls):
...
+    vertical_positions: List[float] = Field(alias="verticalPositions")
+    vertical_interpolation: VerticalInterpolation = Field(alias="verticalInterpolation")
+    vertical_position_type: VerticalPositionType = Field(alias="verticalPositionType")
+
+
+class QHTable(ForcingBase):
+    function: Literal["qhtable"] = "qhtable"
+
+
+class Constant(ForcingBase):
```
BeardedPlatypus, line 117: Technically a constant has only a single value. It might be nice to add a follow up issue to expose it as such.

hydrolib/core/io/bc/models.py
```diff
@@ -136,50 +50,107 @@ def _supports_comments(cls):
...
-
-        data["function"] = function_data
+    def _get_parser(cls) -> Callable:
+        return cls.parse
+
+    @classmethod
+    def parse(cls, filepath: Path):
+        # It's odd to have to disable parsing something as comments
+        # but also need to pass it to the *flattener*.
+        # This method now only supports per model settings, not per section.
```
BeardedPlatypus, line 149: This is by design, such settings are defined per whole file, not per section.

hydrolib/core/io/ini/models.py
```diff
@@ -153,74 +48,126 @@ class Comments(BaseModel, ABC):
...
 
 
-class DataBlockIniBasedModel(IniBasedModel):
-    """DataBlockIniBasedModel defines the base model for ini models with datablocks."""
+class CrosslockGeneral(INIGeneral):
+    fileVersion: str = "3.00"
+    fileType: Literal["crossLoc"] = "crossLoc"
+
+
+class INIModel(FileModel):
```
BeardedPlatypus, line 105: I'm not sure if I understand this class. Is it a model that corresponds with a single ini file? If so, then the doc string does not fit? If the docstring is correct, then shouldn't it use a `.mdu` extension. And shouldn't it need an additional step after the `Parser.parse`?

hydrolib/core/io/ini/models.py
```diff
@@ -153,74 +48,126 @@ class Comments(BaseModel, ABC):
...
             exclude={
                 "start_line",
                 "end_line",
-                "datablock",
                 "content",
             }
         )
 
+
+class INIGeneral(IniBasedModel):
```
BeardedPlatypus, line 86: Shouldn't these and the subsequent classes be in a different file? I feel like they refer to some more specific objects, though I'm not sure which ones.

hydrolib/core/io/ini/parser.py
```diff
@@ -289,3 +290,13 @@ def _is_datarow(self, _: str) -> bool:
         # we assume that we already checked whether it is either a comment,
         # section header or a property
         return self._config.parse_datablocks
+
+    @classmethod
+    def parse(cls, filepath: Path):
+        parser = cls(ParserConfig())
```
BeardedPlatypus, line 296: would you not want to pass the `ParserConfig` with a default parameter? 

```suggestion
    def parse(cls, filepath: Path, config=None):
        if config is None:
            config = ParserConfig()
        parser = cls(config)
```

hydrolib/core/io/mdu/models.py
```diff
@@ -0,0 +1,396 @@
...
+
+
+class Trachytopes(IniBasedModel):
+    TrtRou = "N"
+    TrtDef: Optional[int] = None
+    TrtL: Optional[int] = None
+    DtTrt = 60
+
+
+class Output(IniBasedModel):
```
BeardedPlatypus, line 201: A lot of these values are booleans, but we are assigning them ints, is that we want want to do?

tests/io/test_bc.py
```diff
@@ -1,43 +1,52 @@
...
+    assert forcing.datablock[0] == [1.0, 2.0, 3.0]
 
 
 def test_read_bc_expected_result():
     input_str = inspect.cleandoc(
         """
         [Forcing]
             name              = right01_0001
-            function          = timeSeries
+            function          = timeseries  # what about casing?
```
BeardedPlatypus, line 49: the default we write should be camelCase, but everything is case-insensitive.

Note that this now currently is not a valid bc file, because of the comment.

hydrolib/core/basemodel.py
```diff
@@ -22,7 +31,35 @@ class Config:
...
+
+    def is_file_link(self) -> bool:
+        """Generic attribute for models backed by a file."""
+        return False
+
+    def is_intermediate_link(self) -> bool:
+        """Generic attribute for models that have children fields that could contain files."""
+        return self.is_file_link()
+
+    def show_tree(self, indent=0):
```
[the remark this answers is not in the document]
evetion: `BaseModel` might not be the right name, but this is our (not Pydantics) class that will be used and nested everywhere. Moving it somewhere else would complicate things a bit.

hydrolib/core/io/mdu/models.py
```diff
@@ -0,0 +1,396 @@
...
+    Program: str = "D-Flow FM"
+    Version: str = "1.2.94.66079M"
+    fileType: str = "modelDef"
+    fileVersion: str = "1.09"
+    GuiVersion: str = "1.5.0.49132"
+    AutoStart: bool = False
+    PathsRelativeToParent: bool = False
+
+
+class Numerics(IniBasedModel):
```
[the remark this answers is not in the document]
evetion: Fully agree here. Note that pydantic can create a json schema of all nested classes for documentation.

hydrolib/core/io/structure/models.py
```diff
@@ -38,15 +52,43 @@ class Comments(IniBasedModel.Comments):
...
+    @root_validator
+    def check_location(cls, values):
+        assert (
+            "n_coordinates" in values
+            and "x_coordinates" in values
+            and "y_coordinates" in values
+        ) or (
+            "branchid" in values and "chainage" in values
+        ), "Specify location either by setting `branchid` and `chainage` or `*_coordinates` fields."
+        return values
```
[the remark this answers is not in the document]
evetion: @arthurvd Could you comment on the expected behaviour if both coordinates and branchid are provided?

hydrolib/core/io/structure/models.py
```diff
@@ -60,69 +102,171 @@ class Comments(Structure.Comments):
...
         )
 
-        crest_level: Optional[str] = Field(
+        crestlevel: Optional[str] = Field(
             "Crest level of weir (m AD).", alias="crestLevel"
         )
-        crest_width: Optional[str] = Field("Width of the weir (m).", alias="crestWidth")
-        correction_coefficient: Optional[str] = Field(
+        crestwidth: Optional[str] = Field("Width of the weir (m).", alias="crestWidth")
+        corrcoeff: Optional[str] = Field(
```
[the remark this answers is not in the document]
evetion: Could be added back in. There's no single way now to do these conversions, as the parsers and some of my code also do conversions to remove things like spaces or -. Should be made into an issue.

hydrolib/core/io/structure/models.py
```diff
@@ -60,69 +102,171 @@ class Comments(Structure.Comments):
...
+
+    _split_to_list = get_split_string_on_delimeter_validator("relopening", "losscoeff")
+
+
+class Pump(Structure):
+
+    structure_type: Literal["pump"] = Field("pump", alias="type")
+
+    orientation: str
+    controlside: str = Field(alias="controlSide")  # TODO Enum
```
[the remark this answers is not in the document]
evetion: Agreed.

hydrolib/core/basemodel.py
evetion: Could be, we're now mostly looking for a (child)instance of `FileModel`. Then again, that's unknown to check for on the `BaseModel`, so I thought this was a nice way of doing it and could provide a way to also stop recursively parsing eventually.

hydrolib/core/basemodel.py
evetion: Agreed, that's why we check for `is_absolute` above.

hydrolib/core/io/bc/models.py
evetion: While I agree the naming is super confusing between some files, these class names all derive from the section header.

hydrolib/core/io/bc/models.py
evetion: Well, we might need helper functions for all these classes, although I wouldn't be surprised if this class can represent multiple static values (instead of a series).

hydrolib/core/io/bc/models.py
evetion: This is the documentation of the design, should be followed up by #40

tests/io/test_bc.py
evetion: This was a testcase for the lowercase, hence the comment.

hydrolib/core/io/mdu/models.py
evetion: Should be a followup to document all the fields a bit better and more complete. They're now based on the test files.

hydrolib/core/io/ini/models.py
evetion: This is the header part of the `.ini` files, in which eventual version checks and warnings can be stored. Should be in `models`, like the rest of the representation of these files?

hydrolib/core/io/ini/models.py
evetion: Yeah, single `.ini` model. I've updated the docstring.

PR #45, opened 2021-08-25
Feature/add net tests

BeardedPlatypus:


evetion:
@BeardedPlatypus Could you comment on the state of this PR? Would it be ok to merge after review, or could you make a small todo checklist that someone else could pick up? I don't want these branches to diverge too much.

PR #49, opened 2021-09-13
[WIP] Fixes for model4.mdu

evetion:


hydrolib/core/io/mdu/models.py
```diff
@@ -176,8 +176,10 @@ def is_intermediate_link(self) -> bool:
 class Lateral(INIBasedModel):
     _header: Literal["Lateral"] = "Lateral"
     id: str
-    name: str
-    nodeId: str
+    name: str = ""
+    nodeId: Optional[str]
+    branchId: Optional[str]
+    chainage: Optional[float]
```
priscavdsluis, line 182: I believe a Lateral should have either nodeId OR branchId + chainage. Is it possible to validate this?

hydrolib/core/io/mdu/models.py
```diff
@@ -342,18 +344,29 @@ class Output(INIBasedModel):
...
+    _split_to_list2 = get_split_string_on_delimiter_validator(
+        "hisinterval",
+        "xlsinterval",
+        "mapinterval",
+        "rstinterval",
+        "classmapinterval",
+        "waqinterval",
+        "statsinterval",
+        "timingsinterval",
+        delimiter=" ",
```
priscavdsluis, line 368: I believe setting the delimiter to " " will cause the validation to fail if the parts have multiple whitespaces between them.

Issue #42, opened 2021-07-30
Don't repeatedly parse the same file.


evetion:
For example, in an .ext model, repeated references to the same .bc file are made, which is parsed over and over:

```
         ∟ ExtModel represented by /Users/evetion/code/HYDROLIB-core/tests/data/input/e02/c11_korte-woerden-1d/dimr_model/dflowfm/FlowFM_bnd.ext.
             ∟ ForcingModel represented by /Users/evetion/code/HYDROLIB-core/tests/data/input/e02/c11_korte-woerden-1d/dimr_model/dflowfm/FlowFM_boundaryconditions1d.bc.
             ∟ ForcingModel represented by /Users/evetion/code/HYDROLIB-core/tests/data/input/e02/c11_korte-woerden-1d/dimr_model/dflowfm/FlowFM_boundaryconditions1d.bc.
             ∟ ForcingModel represented by /Users/evetion/code/HYDROLIB-core/tests/data/input/e02/c11_korte-woerden-1d/dimr_model/dflowfm/FlowFM_boundaryconditions1d.bc.
```

PR #61, opened 2021-10-12
Make sure file is only parsed once (#42)

priscavdsluis:


hydrolib/core/basemodel.py
```diff
@@ -132,8 +159,15 @@ def is_file_link(self) -> bool:
     def validate(cls: Type["FileModel"], value: Any):
         # Enable initialization with a Path.
         if isinstance(value, (Path, str)):
+            filepath = Path(value)
+
+            if not filepath.is_absolute():
+                folder = context_dir.get(None)
```
evetion, line 165: This folder could be None right?

hydrolib/core/basemodel.py
```diff
@@ -95,31 +96,57 @@ class FileModel(BaseModel, ABC):
...
-                filepath = folder / filepath
-            # Otherwise we're the root filepath
-            # and should set the context
-            else:
+            if not context_dir.get(None):
                 logger.info(f"Set context to {filepath.parent}")
                 context_dir_reset_token = context_dir.set(filepath.parent)
 
+            if filepath in FileModel._file_models_cache:
+                return
```
evetion, line 141: If you return here (return None is more explicit), context_dir is never reset right?

hydrolib/core/basemodel.py
```diff
@@ -95,31 +96,57 @@ class FileModel(BaseModel, ABC):
...
+        Returns:
+            FileModel: A file model.
+        """
+
+        if filepath:
+            filepath = Path(filepath)
+
+            if filepath in FileModel._file_models_cache:
+                logger.info(
+                    f"Returning existing FileModel from cache, because {filepath} was already parsed"
```
evetion, line 120: You could use class name here instead of FileModel, so users see the actual (sub)class used. You could also add a . to the end of the sentence.

hydrolib/core/basemodel.py
evetion: Might be good to keep the logic for the context_dir in one spot, maybe move the set part to here as well?

hydrolib/core/basemodel.py
```diff
@@ -95,31 +96,57 @@ class FileModel(BaseModel, ABC):
...
+
         context_dir_reset_token = None
         if filepath:
             filepath = Path(filepath)  # so we also accept strings
 
-            # If a context is set, use it
-            if (folder := context_dir.get(None)) and not filepath.is_absolute():
-                logger.info(f"Used context to get {folder} for {filepath}")
-                filepath = folder / filepath
-            # Otherwise we're the root filepath
```
evetion, line 114: I think these comments were useful.

tests/test_model.py
```diff
@@ -195,6 +195,29 @@ def test_mdu_model():
...
+    )
+
+    boundary1 = model.boundary[0]
+    boundary2 = model.boundary[1]
+
+    # Set a field for first boundary
+    boundary1.forcingfile.forcing[0].name = "some_new_value"
+
+    # Field for second boundary is also updated (same instance)
+    assert boundary2.forcingfile.forcing[0].name == "some_new_value"
```
evetion, line 218: 👍🏻

docs/topics/principles.md
```diff
@@ -36,3 +36,40 @@ You can see this tree structure if you call `show_tree`.
            ∟ Boundary
              ∟ ForcingModel represented by FlowFM_boundaryconditions1d.bc.
 ```
+
+# Program flow while recursively creating FileModels
```
evetion, line 40: Nice, this certainly helps.

evetion (changes requested):

hydrolib/core/basemodel.py
priscavdsluis: Done

hydrolib/core/basemodel.py
priscavdsluis: Sorry I lost them in the process somewhere :P. I agree.

evetion (approved):

Issue #57, opened 2021-10-12
Implement RR rainfall file (".bui")


arthurvd:
Support for the rainfall input file for the RR kernel, often called "DEFAULT.BUI":
This file contains the rainfall data: the number of stations, the names (id’s) of the rainfall
stations, the number of rainfall events, the rainfall data timestep size (may be different from
the computation timestep size), the start date and duration of each event and the rainfall data
for each event.

File syntax:
```
*Name of this file: \SOB_LITE\FIXED\DEFAULT.BUI
*Date and time of construction: 17-04-1997 17:13:50
*Enige algemene wenken:
* Gebruik de default dataset voor overige invoer (altijd 1 bij bui, 0 bij reeks)
1
*Aantal stations
1
*Namen van stations
’Station1’
*Aantal gebeurtenissen (omdat het 1 bui betreft is dit altijd 1)
*en het aantal seconden per waarnemingstijdstap (10800 = 3x3600)
1 10800
*Elke commentaarregel wordt begonnen met een * (asteriks).
*Eerste record bevat startdatum en -tijd, lengte van de gebeurtenis in dd hh mm ss
*Het format is: yyyymmdd:hhmmss:ddhhmmss
*Daarna voor elk station de neerslag in mm per tijdstap.
1996 1 1 0 0 0 1 3 0 0
0.2
0.2
0.2
0.2
0.2
0.2
0.2
0.2
0.2
```
(from: https://content.oss.deltares.nl/delft3d/manuals/SOBEK_User_Manual.pdf#subsection.D.19.3, page 819)

Note that @d2hydro already has a prototype implementation in HYDROLIB/contrib/d2hydro/modules/stowa_bui_generator/src/stowabui/stowa_bui_generator.py

Under this issue create a parser/class/serializer in the standard approach as done for other file types in HYDROLIB-core.

**Regarding multiple events (.rks files, buienreeksen)**
Quote from Geert Prinsen:

> In the SOBEK2 GUI the extension .BUI was always used for one single 1 bui  (rainfall event), (so the default.bui file contains 1 bui), and the extension .RKS for a series of events. Still, for the computational cores, the filename extension **does not matter**, based on the contents of the file (header), the file is processed and used either as a single event, or as a series of multiple events.(since the file header contains the number of rainfall events).

PR #64, opened 2021-10-14
#57 Implement RR rainfall file (".bui") 

Carsopre:
Now we can parse correctly .bui files.

- It should be able to parse / serialize models with multiple precipitation events as well as multiple stations (tests still needed as no test data is yet available).
- Followed the same structure models/serializer/parser as in the other models (for consistency).
- Added tests in the same fashion as the rest.


hydrolib/core/io/bui/models.py
```diff
@@ -0,0 +1,50 @@
...
+from hydrolib.core.basemodel import BaseModel, FileModel
+from .parser import BuiParser
+from .serializer import write_bui_file
+
+class BuiPrecipitationEvent(BaseModel):
+    start_time: datetime
+    timeseries_length: timedelta
+    precipitation_per_timestep: List[List[float]]
+
+class BuiPrecipitationEventList(BaseModel):
```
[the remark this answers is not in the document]
Carsopre: There can be more than one precipitation event, I already discussed it with Arthur and that's why i'm waiting until a test file can be used to verify it's correct functioning. Also, my (short) experience with pydantic taught me is better dividing and conquer, resulting in this single class containing a single property to represent the list of events.

hydrolib/core/io/bui/parser.py
```diff
@@ -0,0 +1,133 @@
...
+    StartTime (YYYY mm dd HH MM SS) TimeSeriesLength (dd HH MM SS)
+    PrecipitationPerTimestep
+    StartTime (YYYY mm dd HH MM SS) TimeSeriesLength (dd HH MM SS)
+    PrecipitationPerTimestep
+    Example given:
+    2021 12 20 0 0 0 1 0 4 20
+    4.2
+    4.2
+    4.2
+    2021 12 21 0 0 0 1 0 4 20
```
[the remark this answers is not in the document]
Carsopre: Based on a talk with Arthur your statement is not correct (I refactored my code from yesterday after a talk with Arthur based on this point).

hydrolib/core/io/bui/models.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @classmethod
+    def _get_serializer(cls) -> Callable:
+        return write_bui_file
+
+    @classmethod
+    def _get_parser(cls) -> Callable:
+        return BuiParser.parse
+
+    def get_station_events(self, station: str) -> Dict[datetime, List[float]]:
```
evetion, line 78: Would be good to document this in the documentation with a tutorial, or make sure it shows up in the API reference. (add bui.md file in docs/reference).

hydrolib/core/io/bui/parser.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    A parser for the precipitation event section within a .bui file.
+    It resembles something like this:
+    StartTime (YYYY mm dd HH MM SS) TimeSeriesLength (dd HH MM SS)
+    PrecipitationPerTimestep
+    Example given:
+    2021 12 20 0 0 0 1 0 4 20
+    4.2 2.4
+    4.2 2.4
+    4.2 2.4
```
evetion, line 16: 42 is 👍🏻

hydrolib/core/io/fnm/models.py
```diff
@@ -28,7 +29,7 @@ class RainfallRunoffModel(FileModel):
     pluvius: Path = Path("pluvius.3b")
     pluvius_general: Path = Path("pluvius.alg")
     kasklasse: Path = Path("kasklass")
-    bui_file: Path = Path("default.bui")
+    bui_file: Optional[BuiModel] = None
```
evetion, line 32: Might need to check whether these fields (in the .fnm.file) always need to show the default filename, whether it exists or not. Some test changes are related to this.

hydrolib/core/io/bui/serializer.py
```diff
@@ -0,0 +1,217 @@
...
+
+def write_bui_file(path: Path, data: Dict) -> None:
+    """
+    Writes a .bui file in the given path based on the data given in a dictionary.
+
+    Args:
+        path (Path): Path where to output the text.
+        data (Dict): Data to serialize into the file.
+    """
+    data["filepath"] = path  # This is redundant as already exists in the data.
```
evetion, line 213: If it's redundant, why not leave it out? Serializer won't do anything with it?

hydrolib/core/io/fnm/models.py
Carsopre: From my understanding if a default path is given but it does not exist it will try to create the model based on the current logic, so it will 'crash'. If no path is given then nothing should happen.

Issue #58, opened 2021-10-12
Make sure correct information (by id) is taken from the Boundary bc file (instead of the whole file)


priscavdsluis:
The external forcing file (.ext) may contain boundaries that reference a boundary condition file (.bc).
Not all the data in the .bc file is relevant for this one boundary; only one section is relevant.
This section can be coupled by using the boundary name and quantity.

## Example:
**External forcing file**

```
[boundary]
quantity    = waterlevelbnd
nodeId      = bnd1
forcingfile = BoundaryConditions.bc

[boundary]
quantity    = dischargebnd
nodeId      = bnd2
forcingfile = BoundaryConditions.bc

[boundary]
quantity    = qhbnd
nodeId      = bnd3
forcingfile = BoundaryConditions.bc
```

**Boundary conditions file**
```
[forcing]
    name                  = bnd1          
    function              = constant            
    time-interpolation    = linear  
    quantity              = waterlevelbnd         
    unit                  = m                   
    2.5 

[forcing]
    name                  = bnd2           
    function              = constant            
    time-interpolation    = linear  
    quantity              = dischargebnd     
    unit                  = m³/s                
    100

[forcing]
    name                  = bnd3          
    function              = qhtable
    quantity              = qhbnd discharge
    unit                  = m³/s
    quantity              = qhbnd waterlevel
    unit                  = m
    50 1.25
    100 2.5
    150 3.75  
```

PR #65, opened 2021-10-15
Make sure user can retrieve the correct boundary information from the bc file (#58)

priscavdsluis:
… bc file. (#58)


tests/test_model.py
```diff
@@ -200,3 +201,45 @@ def test_mdu_from_scratch():
     model = FMModel()
     model.filepath = output_fn
     model.save()
+
+
+def test_boundary_with_forcingfile_returns_forcing():
```
Carsopre, line 206: What is the purpose of having three boundaries on the same test? They are (virtually) doing the same. I believe with only one boundary should suffice.

PR #69, opened 2021-10-19
Feature/54 validation in laterals

Carsopre:
Added validation based on the following conditions:

1. When providing either x or y coordinates, the size of the values given should match the one specified in numCoordinates.
2. When specifying a locationType:
    2.1. Raise ValueError If either x or y coordinates do not contain values.
    2.2. Raise ValueError If the locationType string value is not "1d", "2d" or "all".
    2.3. If the locationType is "1d" and:
      2.3.1. Succeed if nodeId field has been given and is valid.
      2.3.2. Raise ValueError if nodeId has not been given and branchId and chainage values are not valid.

In addition, tests created under tests/io/test_mdu.py


tests/test_utils.py
```diff
@@ -15,3 +16,17 @@ def test_split_string_strip_semicolon(self):
...
+    @pytest.mark.parametrize(
+        "input_str",
+        [
+            pytest.param("", id="No spaces"),
+            pytest.param(" ", id="Spaces"),
+            pytest.param(None, id="None"),
+            pytest.param("\t", id="Tabulator"),
+        ],
+    )
+    def test_given_args_expected_result(self, input_str: str):
```
priscavdsluis, line 31: Would add a test for the case where it returns false.

hydrolib/core/io/mdu/models.py
```diff
@@ -177,11 +182,119 @@ class Lateral(INIBasedModel):
     _header: Literal["Lateral"] = "Lateral"
     id: str
     name: str = ""
+    locationType: Optional[str]
     nodeId: Optional[str]
     branchId: Optional[str]
     chainage: Optional[float]
+    numCoordinates: Optional[int]
+    xCoordinates: Optional[List[int]]
+    yCoordinates: Optional[List[int]]
```
priscavdsluis, line 191: I believe that if there are one or multiple coordinates in the file, they still need to be split into a list, otherwise pydantic cannot make the conversion. You could take a look at the usages of the `get_split_string_on_delimiter_validator`.

hydrolib/core/utils.py
```diff
@@ -32,3 +32,16 @@ def to_list(item: Any) -> List[Any]:
     if not isinstance(item, list):
         return [item]
     return item
+
+
+def str_is_empty_or_none(str_field: str) -> bool:
```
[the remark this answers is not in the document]
priscavdsluis: Thanks!

Sorry for being unclear, I was trying to say the documentation should also be updated to describe that it also checks for a string being all white space (in addition to being empty or none)

hydrolib/core/io/mdu/models.py
priscavdsluis: The validator is tested through for e.g. the test `test_split_string_strip_whitespace`.

priscavdsluis (approved):

Issue #71, opened 2021-10-19
Validation in Boundary Conditions


Carsopre:
**- Boundary conditions: https://content.oss.deltares.nl/delft3d/manuals/D-Flow_FM_User_Manual_1D2D.pdf#subsubsection.C.5.2.1
(either nodeId, or locationFile)**

_Originally posted by @arthurvd in https://github.com/Deltares/HYDROLIB-core/issues/54#issuecomment-945546926_

PR #83, opened 2021-10-20
#71 validation in boundary conditions

Carsopre:


hydrolib/core/io/mdu/models.py
```diff
@@ -168,13 +166,37 @@ class Restart(INIBasedModel):
...
+            ValueError: When none of the values are present.
+
+        Returns:
+            Dict: Validated dictionary of values for Boundary.
+        """
+        node_id = values.get("nodeid", None)
+        forcingfile = values.get("forcingfile", None)
+        if str_is_empty_or_none(node_id) and not isinstance(forcingfile, ForcingModel):
+            raise ValueError(
+                "Either nodeid or locationfile fields should be specified."
```
arthurvd, line 196: Please use correct casing in end user messages, so in this case: "Either nodeId or locationFile fields should be specified."

hydrolib/core/io/mdu/models.py
```diff
@@ -168,13 +166,37 @@ class Restart(INIBasedModel):
...
+
+        Raises:
+            ValueError: When none of the values are present.
+
+        Returns:
+            Dict: Validated dictionary of values for Boundary.
+        """
+        node_id = values.get("nodeid", None)
+        forcingfile = values.get("forcingfile", None)
+        if str_is_empty_or_none(node_id) and not isinstance(forcingfile, ForcingModel):
```
arthurvd, line 194: This condition seems to check on presence of either nodeid, or forcingfile, but I think it should check on either nodeid or locationFile. (forcingFile is always required!)

hydrolib/core/io/mdu/models.py
```diff
@@ -168,13 +166,37 @@ class Restart(INIBasedModel):
 class Boundary(INIBasedModel):
     _header: Literal["Boundary"] = "Boundary"
     quantity: str
-    nodeId: Optional[str]
+    nodeid: Optional[str]
+    locationfile: Optional[Path]
     forcingfile: Optional[ForcingModel] = None
     bndWidth1D: Optional[float]
```
arthurvd, line 172: This is correct, but another field is missing:
bndBlDepth: Optional[float]

tests/io/test_mdu.py
```diff
@@ -386,3 +388,113 @@ def test_given_valid_args_validates_locationtype(self, location_dict: str):
                 assert isinstance(lateral_cls, INIBasedModel)
                 for key, value in lateral_dict.items():
                     assert lateral_cls.dict()[key] == value
+
+    class TestBoundary:
```
arthurvd, line 392: @priscavdsluis : can you review the unit tests below?

arthurvd (changes requested):
See my individual review comments.

hydrolib/core/io/mdu/models.py
```diff
@@ -1,11 +1,9 @@
 from pathlib import Path
-from typing import Callable, Dict, List, Literal, Optional, Union
+from typing import Dict, List, Literal, Optional, Union
 
-from pydantic import Field, validator
-from pydantic.class_validators import root_validator
+from pydantic import Field, root_validator, validator
+from pydantic.error_wrappers import ValidationError
```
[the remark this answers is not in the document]
evetion: If the imports work, I'm fine with it. I do even think that all can be imported from pydantic itself.

hydrolib/core/io/mdu/models.py
Carsopre: in the same model these properties are defined in lower case.

hydrolib/core/io/mdu/models.py
```diff
@@ -168,13 +166,37 @@ class Restart(INIBasedModel):
 class Boundary(INIBasedModel):
     _header: Literal["Boundary"] = "Boundary"
     quantity: str
-    nodeId: Optional[str]
+    nodeid: Optional[str]
+    locationfile: Optional[Path]
     forcingfile: Optional[ForcingModel] = None
```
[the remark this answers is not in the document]
Carsopre: Fixed.

hydrolib/core/io/mdu/models.py
Carsopre: Modified properties names and messages (also in the tests).

arthurvd (approved):

Issue #86, opened 2021-10-21
Test test_mdu_from_scratch() has KeyError 'filepath'


arthurvd:
Local run of unit tests gives error: def test_mdu_from_scratch(): // KeyError: 'filepath'

Part of Stack trace:
```
KeyError: 'filepath'
self = FMModel(), data = {}

    def __init__(self, **data: Any) -> None:
        """Initializes a BaseModel with the provided data.
    
        Raises:
            ValidationError: A validation error when the data is invalid.
        """
        try:
>           super().__init__(**data)

hydrolib\core\basemodel.py:53: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[..]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = FMModel(), data = {}

    def _get_identifier(self, data: dict) -> str:
>       return data["filepath"].name
E       KeyError: 'filepath'

hydrolib\core\basemodel.py:274: KeyError
```

PR #88, opened 2021-10-21
Make sure that no Key exception is thrown when retrieving the file pa…

priscavdsluis:
…th (#86)


arthurvd (commented):
See individual comment

PR #122, opened 2021-10-31
Support hashed string values in ini properties

### LINKED ISSUE #121: Support #enclosed# string values in ini files
```
[Section]
keyA = #valueA#   # Some comment
keyB = valueB#Some comment
```

arthurvd:
Fixes #121

### DIFF
--- a/hydrolib/core/io/ini/parser.py
+++ b/hydrolib/core/io/ini/parser.py
@@ -239,7 +239,6 @@ class Parser:
 
     def _retrieve_property_comment(self, line: str) -> Tuple[Optional[str], str]:
         if self._config.parse_comments and self._config.comment_delimiter in line:
-            # key_value, comment = line.split(self._config.comment_delimiter, 1)
             parts = line.strip().split(self._config.comment_delimiter)
             numhash = len(parts) - 1
             if numhash == 1:
--- a/hydrolib/core/io/ini/parser.py
+++ b/hydrolib/core/io/ini/parser.py
@@ -238,6 +238,30 @@ class Parser:
         self._current_section.add_datarow(line.split())  # type: ignore
 
     def _retrieve_property_comment(self, line: str) -> Tuple[Optional[str], str]:
+        """Retrieve the comment and value part from the valuestring of a key-value pair.
+
+        The comment retrieval is complicated by the fact that in the Deltares-INI
+        dialect, the comment delimiter '#' plays a double role: it may also be used
+        to quote string values (for example if the contain spaces).
+
+        Example lines that are supported:
+        key = valueAndNoComment
+        key = valueA  # and a simple comment
+        key = #valueA with possible spaces#
+        key = #valueA#  # and a simple comment
+        key = #valueA# # and a complicated comment with hashes #1 example
+        key = value # and a complicated comment with hashes #2.
+
+        Keywords arguments:
+            line (str) -- the partial string of the line containing both value and
+                possibly a comment at the end. Note that the "key =" part must already
+                have been split off, for example by _retrieve_key_value
+
+        Returns:
+            Tuple with the comment and string value, respectively. If no comment is
+            present, the first tuple element is None.
+        """
+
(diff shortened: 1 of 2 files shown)


hydrolib/core/io/ini/parser.py
```diff
@@ -239,8 +239,26 @@ def _handle_datarow(self, line: str) -> None:
 
     def _retrieve_property_comment(self, line: str) -> Tuple[Optional[str], str]:
         if self._config.parse_comments and self._config.comment_delimiter in line:
-            key_value, comment = line.split(self._config.comment_delimiter, 1)
-            return comment.strip(), key_value.strip()
+            # key_value, comment = line.split(self._config.comment_delimiter, 1)
```
evetion, line 242: Remove comment

hydrolib/core/io/ini/parser.py
```diff
@@ -239,8 +239,26 @@ def _handle_datarow(self, line: str) -> None:
...
+            parts = line.strip().split(self._config.comment_delimiter)
+            numhash = len(parts) - 1
```
evetion, line 244: Might be better to just use `line.count("#")`, and only split/strip when really required.

hydrolib/core/io/ini/parser.py
```diff
@@ -239,8 +239,26 @@ def _handle_datarow(self, line: str) -> None:
...
+            if numhash == 1:
+                # normal value, simple comment: "key =  somevalue # and a comment "
+                return parts[-1].strip(), parts[0].strip()
+            elif parts[0] == "":
```
evetion, line 248: This logic is odd to read. Might be better to write `numhash == 2` (case without comment).

hydrolib/core/io/ini/parser.py
```diff
@@ -239,8 +239,26 @@ def _handle_datarow(self, line: str) -> None:
...
+                # hashed value, possible with comment: "key = #somevalue# ..."
+                return (
+                    self._config.comment_delimiter.join(parts[3:]).strip()
+                    if numhash >= 3
+                    else None,
+                    self._config.comment_delimiter.join(parts[0:3]).strip(),
+                )
+            else:
```
evetion, line 256: With the other comments, this automatically becomes 3 or more hashes. So a hashed value and a comment (which may or not contain more hashes).

evetion (changes requested):
Maybe also add a docstring/larger commented part that details the three cases you can have in terms of comments (the why of this function) together, not just in between the lines.

hydrolib/core/io/ini/parser.py
arthurvd: It is deliberate: it catches these cases:
```
key = #valueA# # and a comment
key = #valueA with spaces # # Comment: and give me your #1 reason for writing such a comment?
```
if the first token of split is empty string (and numhas > 1), then we have a starting "#", which must always mean hash-string-quoting.

hydrolib/core/io/ini/parser.py
arthurvd: is already done by existing if in line above: `[..]and self._config.comment_delimiter in line:`

hydrolib/core/io/ini/parser.py
arthurvd: It catches this silly case:
```
key = valueA # and here is hash #2, funny huh?
```

hydrolib/core/io/ini/parser.py
arthurvd: Please check whether with my explanations above any of this changes are still needed.

arthurvd:
> Maybe also add a docstring/larger commented part that details the three cases you can have in terms of comments (the why of this function) together, not just in between the lines.

See latest commit 8d33250.

hydrolib/core/io/ini/parser.py
arthurvd: See latest commit

Issue #101, opened 2021-10-22
Saving .bc files incorrectly writes repeated key names as a semicolon-separated list.


arthurvd:
When loading an MDU with underlying ext + bc files, and then subsequently saving the model again, a wrong .bc file is produced for repeated fieldnames.
Example valid input:
```
[Forcing]
name       = Weir_RS373-st1
function   = timeseries
timeInterpolation = linear
quantity   = time
unit       = minutes since 2000-01-01 00:00:00
quantity   = weir_crestLevel
unit       = m+NAP
    0.000000    10.429000 
   60.000000    10.429000 
```

Resulting wrong serialized bc file:
```
[forcing]
    name              = Weir_RS373-st1
    function          = timeseries
    quantity          = time;weir_crestLevel  # <<< WRONG! Should *not* be a ;-separated list, but multiple lines.
    unit              = minutes since 2000-01-01 00:00:00;m+NAP  # <<< WRONG! Should *not* be a ;-separated list, but multiple lines.
    timeinterpolation = linear
    offset            = 0.0
    factor            = 1.0
        0.0        10.429
        60.0       10.429
```

Note this may be due to MDU files supporting *unique* field names with multiple values a semicolon-separated list.
The .bc file is the *only* input file, which is INI-like, but does not use the same readers/writers as all other structure/MDU/ext ini-like files.

PR #134, opened 2021-11-29
#101: Duplicate keys in bc file incorrectly leads to semicolon separated keys

priscavdsluis:


evetion:
Interesting, there seem to be unicode issue on Windows

hydrolib/core/io/bc/models.py
```diff
@@ -37,15 +33,24 @@ class TimeInterpolation(str, Enum):
     block_to = "blockTo"
 
 
+class QuantityUnitPair(NamedTuple):
+    quantity: str
+    unit: str
+
+    def to_properties(self):
```
evetion, line 40: Could be `_to_properties`?

evetion (approved):
Nice code. Could you provide a test example, or documentation example where you intialize the quantities with keys? Now it could be unclear, I rather have:

```python
       quantities=[
            (quantity="qhbnd discharge", unit="m3/s"),
            (quantity="qhbnd waterlevel", unit="m"),
        ],
```

Than the unnamed, current, solution:
```python
       quantities=[
            ("qhbnd discharge", "m3/s"),
            ("qhbnd waterlevel", "m"),
        ],
```

tests/io/test_bc.py
```diff
@@ -98,7 +106,7 @@ class TestForcingBase:
         ],
     )
     def test_parses_function_case_insensitive(self, input, expected):
-        forcing = ForcingBase(function=input, name="somename", quantity=[], unit=[])
+        forcing = ForcingBase(function=input, name="somename", quantities=[])
```
arthurvd, line 109: In some earlier commit, apparently ```quantity``` was renamed to ```quantities```.
This intuitive, but it is inconsistent with class members in many other classes, for example: ```StructureModel.structure```, ```CrossDefModel.definition```, etc., which are all also lists.
Not sure what is best to change or keep here...

arthurvd (approved):

tests/io/test_bc.py
priscavdsluis: Changed the name as agreed to `quantityunitpair`

arthurvd (approved):

PR #136, opened 2021-11-30
Do not write absolute file paths to file

### LINKED ISSUE #96: Do not write absolute file paths to file
I noticed that for example in the mdu file we reference other files with their full file path. 
It might have been caused by issue #42 if this was not previously an issue.

BeardedPlatypus:
Fixes #96


evetion:
Based on all our discussions:


`save(filepath=None, recurse=False)` (default)
	- Save in it's (absolute) original place, or filepath is given (which then mutates)
	- filepath can also be set directly beforehand (!)

`save(filepath=None, recurse=True)`
	- Save itself and all its children
	- Children are saved relative to parent (can result in weird layout if self is not the original root)
	- Absolute paths do not change
	- If filepath is given, it mutates the state beforehand

`export(folder)`
	- non mutating move, self contained
	- overrules relative/absolute paths
	- requires default relative paths

Some thought need to be given to how relative/absolute filepaths work, and when these are mutated.

evetion (approved):
I like it so far, it's huge, but there are clear improvements. Can you comment on the replacement of _apply_recurse and is_intermediate_link on the `BaseModel`?

BeardedPlatypus:
> 

I have created issue #170 to address the `export` functionality. Given the size of this issue I believe this should not be done now. 

There are some functionality choices which need to be made (@arthurvd), with the chosen approach:

* When do we set the internal `absolute_filepath`?
* Given a model with some relative `filepath` and some previously saved `absolute_filepath` how do we resolve the relative filepath?

Given the description by @evetion, I believe the best course of action would be to have a private attribute for the last read absolute path. We can then use this path to resolve some of the relative paths. This does however require some choices to be made, both stemming from the fact that we do not know where the models are in the `FileModel` hierarchy:

## When do we synchronise/store the absolute file paths of all the models.

We resolve the absolute filepaths when we first initialise a ` FileModel` and when we perform a recursive save. We could update the `filepath` field of all `FileModel` objects in the hierarchy at these two points. However, it is impossible to know, given the current structure, whether the entry point is a child in the model hierarchy, or the root. 

As such this could potentially lead to weird behaviour:

If we are dealing with a FMModel with `pathsrelativetoparent` set to `False`, we can no longer assume the paths in a sub model are relative to the direct parent. As such, saving a sub model like the external forcings might result in the forcing files saved in a wrong directory.  Furthermore, if we subsequently store these forcings file, they will also be stored in these wrong locations (as the absolute path was updated during the recursive save of the external forcings file).

## How do we resolve the paths stored in the `filepath` of a (sub)model.

There are two problems regarding resolving the paths stored in the `filepath`:

* How do we resolve a relative path in respect to the absolute path
* We cannot save the actual root model with a relative path 

Because we do not know whether a model is the root model or a sub model, if a model has a relative path, we need to use the stored absolute path to resolve it. This means that we cannot save a root model with a relative path in regards to the current working directory anymore. As such, it is the users responsibility to have set an absolute path for the root model. This needs to be explicitly documented

If we do encounter a relative path, for example when we save a sub model, we need to resolve it. There are a few options for this:

1. We always use the stored internal absolute path, ignoring any changes to the filepath in the model
2. We resolve the relative path by appending it to the parent of the absolute path
3. We try and find the respective root from which we resolve by verifying if the first parts of the relative path match the last parts of the absolute path

These have respectively the following problems:

1. Changes to the `filepath` are ignored, and the user has no way of controlling where we exactly save, unless the `filepath` is set to some absolute path.
2. If we have the relative path `./ext/forcings.ext` which might be saved at some location `D:\model\ext\forcings.ext`, calling save on this model would lead to the external forcings file being saved in `D:\model\ext\ext\forcings.ext`
3. The previous problem would not occur, however if we were to change the sub folder, e.g. to `./ext_other/forcings.ext`, the model would still be saved in the `D:\model\ext\ext_other\forcings.ext` location. 

This requires a functional choice, which needs to be clearly documented.

hydrolib/core/io/mdu/models.py
```diff
@@ -416,3 +417,13 @@ def _ext(cls) -> str:
     @classmethod
     def _filename(cls) -> str:
         return "fm"
+
+    @FileModel._relative_mode.getter
```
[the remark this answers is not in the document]
BeardedPlatypus: I have added additional information to both the `ResolveRelativeMode` as well as this particular override.

hydrolib/core/basemodel.py
```diff
@@ -523,6 +563,20 @@ def _resolved_filepath(self) -> Optional[Path]:
...
             return context.resolve(self.filepath)
 
+    @property
+    def save_location(self) -> Path:
+        """Gets the current save location which will be used when calling `save()`
+
+        Returns:
+            Path: The location at which this model will be saved.
+        """
+        filepath = self.filepath or self._generate_name()
```
evetion, line 573: Nitpick, this behaviour is a duplicate of the actual save functionality. If you cannot easily change this, maybe add a comment in the save function to always update this one as well.

evetion (approved):

hydrolib/core/basemodel.py
BeardedPlatypus: It is not really a duplicate, we can call the `save_location` from a model created in memory, in which case it seems weird to change state when calling a property. I'd honestly rather have a default value for filepath, and not have it optional. This way we do not have to worry about it being potentially `None` anymore. 

If you agree I will create a follow up issue to do this.

Issue #78, opened 2021-10-19
Improve validation of optionally required fields in orifice structure


arthurvd:
Validation of a structure file shows that some optional fields, and some optionally required fields are now validated too strictly for orifice structures.
More specifically:
- Field `allowedFlowDir` is optional, default is "both".
- Field `useLimitFlowPos` is optional, default is false.
- Field `limitFlowPos` is optionally required, required when `useLimitFlowPos` equals 1/true
- Field `useLimitFlowNeg` is optional, default is false.
- Field `limitFlowNeg` is optionally required, required when `useLimitFlowNeg` equals 1/true

PR #142, opened 2021-12-01
#78: Improve validation of optionally required fields in orifice structure

priscavdsluis:


tests/io/test_structure.py
```diff
@@ -1397,3 +1397,48 @@ def test_given_invalid_values_raises_expectation(
...
+
+
+class TestOrifice:
+    """
+    Wrapper class to test all the methods and sublcasses in:
+    hydrolib.core.io.structure.models.py Orifice class.
+    """
+
+    def test_create_orifice(self):
+        structure = Orifice(**self._create_orifice_values())
```
arthurvd, line 1388: Would it not be nicer to do something like this (to avoid the code duplication for all values):
```
orifdict = **self._create_orifice_values()
structure = Orifice(orifdict)
for key, value in orifdict.items():
    assert get_attr(structure, key) == value

**Review by arthurvd — APPROVED** [review]

File: tests/io/test_structure.py (hunk 1 above)
Comment (priscavdsluis) [reply tests/io/test_structure.py:1389]: As agreed, will not do this. I think it is easier to analyze when the test fails with the current solution.

File: tests/io/test_structure.py [hunk 2]
```diff
@@ -1443,21 +1444,39 @@ def test_orifice_parses_flowdirection_case_insensitive(self, input, expected):
...
-            allowedflowdir="positive",
             crestlevel="2.34",
-            crestwidth="3.45",
             gateloweredgelevel="4.56",
             corrcoeff="5.67",
             usevelocityheight="true",
+        )
+
+    def _create_orifice_values(self) -> dict:
+        optional_values = dict(
```
arthurvd, line 1471: confusing name ```optional_values```, because in the lines below the required ones are also added.

arthurvd (changes requested):
@priscavdsluis : only a small remark, most looks fine!

tests/io/test_structure.py
priscavdsluis: Have changed it to orifice_values

arthurvd (approved):

PR #145, opened 2021-12-02
Feature/144 crsdef tests

### LINKED ISSUE #124: Add unit tests for all CrossSectionDefinition subclasses
Followup to #123

arthurvd:
Fixes #124.

Any idea on how to reduce copy-pasted code in my commits?
I did split off a reusable function CrossSectionDefinition._validate_friction_specification(), but the check_friction validators in CircleCrsDef, RectangleCrsDef and ZwRiverCrsDef are really copy-pasted, and the unit tests in test_crosssection.py are also really quite similar.
I still have three more subclasses to go...

### DIFF
--- a/hydrolib/core/io/ini/models.py
+++ b/hydrolib/core/io/ini/models.py
@@ -25,7 +25,7 @@ from hydrolib.core.io.base import DummySerializer
 from hydrolib.core.io.ini.util import (
     get_from_subclass_defaults,
     get_split_string_on_delimiter_validator,
-    make_list_length_validator,
+    make_list_length_root_validator,
 )
 
 from .io_models import (
@@ -515,19 +515,13 @@ class ZWRiverCrsDef(CrossSectionDefinition):
         "frictionids", "frictiontypes", "frictionvalues"
     )
 
-    @root_validator
-    @classmethod
-    def check_lists(cls, values: dict) -> dict:
-        _check_list_length = make_list_length_validator(
-            "levels",
-            "flowwidths",
-            "totalwidths",
-            length=values["numlevels"],
-            length_name="numlevels",
-            annotation=f"id = {values['id']}",
-        )
-
-        return values
+    _check_list_length = make_list_length_root_validator(
+        "levels",
+        "flowwidths",
+        "totalwidths",
+        length_name="numlevels",
+        annotation="id",
+    )
 
 
 class ZWCrsDef(CrossSectionDefinition):
--- a/hydrolib/core/io/ini/models.py
+++ b/hydrolib/core/io/ini/models.py
@@ -16,6 +16,7 @@ from typing import (
     Union,
 )
 
+from _typeshed import Self
 from pydantic import Extra, Field, root_validator
 from pydantic.class_validators import validator
 
@@ -290,6 +291,52 @@ class CrossSectionDefinition(INIBasedModel):
                 )
         return super().validate(v)
 
+    @classmethod
+    def _validate_friction_specification(
+        cls,
+        values: dict,
+        frictionid_attr: str,
+        frictiontype_attr: str,
+        frictionvalue_attr: str,
+    ) -> bool:
+        """
+        Verifies whether the crosssection definition (subclass) has a valid friction specification.
+        Supposed to be called from with subclass validators for their friction fields.
+
+        Args:
+            values (dict): Dictionary of values to create a CrossSectionDefinition subclass.
+            frictionid_attr: name of the frictionid attribute in the subclass.
+            frictiontype_attr: name of the frictiontype attribute in the subclass.
+            frictionvalue_attr: name of the frictionvalue attribute in the subclass.
+
+        Raises:
+            ValueError: When the values dictionary does not contain valid friction keys.
+
+        Returns:
+            bool: True when input is valid, False otherwise.
+        """
+        frictionid = getattr(cls, frictionid_attr, "")
+        frictiontype = getattr(cls, frictiontype_attr, "")
+        frictionvalue = getattr(cls, frictionvalue_attr, "")
+
+        if frictionid == "":
+            if frictiontype == "" or frictionvalue == "":
+                raise ValueError(
+                    f"Cross section with id \"{values.get('id', '')}\" is missing any friction specification."
+                )
+        else:
+            if frictiontype != "" or frictionvalue != "":
+                raise ValueError(
+                    f"Cross section with id \"{values.get('id', '')}\" has duplicate friction specification (both {frictionid_attr} and {frictiontype_attr}/{frictionvalue_attr})."
+                )
+
+        return True
+
+        # if CrossSectionDefinition._validate_friction_specification(
+        #    cls, values, "frictionid", "frictiontype", "frictionvalue"
+        # ):
+        #    return values
+
 
 class CrossDefModel(INIModel):
     general: CrossdefGeneral = CrossdefGeneral()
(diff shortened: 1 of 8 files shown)


priscavdsluis:
@arthurvd, with regards to the duplicity I think you could do something like the following:

- Define a method that return the root validator:
`def get_friction_validator(
        frictionid_attr: str,
        frictiontype_attr: str,
        frictionvalue_attr: str,
    ):

        def validate_friction_specification(cls, values):
            frictionid = values.get(frictionid_attr) or ""
            frictiontype = values.get(frictiontype_attr) or ""
            frictionvalue = values.get(frictionvalue_attr) or ""

            if frictionid == "":
                if frictiontype == "" or frictionvalue == "":
                    raise ValueError(
                        f"Cross section with id \"{values.get('id', '')}\" is missing any friction specification."
                    )
            else:
                if frictiontype != "" or frictionvalue != "":
                    raise ValueError(
                        f"Cross section with id \"{values.get('id', '')}\" has duplicate friction specification (both {frictionid_attr} and {frictiontype_attr}/{frictionvalue_attr})."
                    )

            return True

        return root_validator(allow_reuse=True)(validate_friction_specification)`

- Then use this validator by doing the following in the body of the class that you want to use it in:
`_friction_validator = get_friction_validator("frictionid", "frictiontype", "frictionvalue")`

arthurvd:
@evetion : Please check my intermediate commit 180b765.
My goal is to use elegant and few lines of code to have validators for fields ```ZwRiverCrsDef.levels```, ```.flowwidths```, and ```.totalwidths```.
Unit test ```test_create_a_zwrivercrsdef_from_scratch()``` now deliberately contains an error input, but not validation is yet triggered about wrong list length. Only the unit test's assert kicks in, but that is too late.

Since I need other items in the values dict, I was going for the root_validator, but can't get it to work.

arthurvd:
> * _friction_validator = get_friction_validator("frictionid", "frictiontype", "frictionvalue")

@priscavdsluis: it worked! After some minor changes, in d516034. Thanks.

hydrolib/core/io/ini/models.py
```diff
@@ -285,6 +286,53 @@ def validate(cls, v):
...
+            """
+            The actual validator function.
+
+            Args:
+            cls: The subclass for which the root_validator is called.
+            values (dict): Dictionary of values to create a CrossSectionDefinition subclass.
+            """
+            frictionid = values.get(frictionid_attr) or ""
+            frictiontype = values.get(frictiontype_attr) or ""
+            frictionvalue = values.get(frictionvalue_attr) or ""
```
priscavdsluis, line 319: You _could_ also remove the ` or ""` and check for None, but it does not really matter :)

hydrolib/core/io/ini/models.py
```diff
@@ -285,6 +286,53 @@ def validate(cls, v):
...
+                    raise ValueError(
+                        f"Cross section with id \"{values.get('id', '')}\" is missing any friction specification."
+                    )
+            else:
+                if frictiontype != "" or frictionvalue != "":
+                    raise ValueError(
+                        f"Cross section with id \"{values.get('id', '')}\" has duplicate friction specification (both {frictionid_attr} and {frictiontype_attr}/{frictionvalue_attr})."
+                    )
+
+            return values
```
priscavdsluis, line 332: It is good practice to override the `BaseModel_get_identifier` method for each identifiable derived BaseModel class. This way, when **any** ValidationError is raised for this class, the identifier will be included in the resulting error message, which makes it easier for the user to identify problems with their model. 
In this case you have added the id in the error message for this specific validation, which is good, but for other validation errors the id will not be included. 
Please check for example `Boundary._get_identifier` in hydrolib.core.io.ext.models.py

hydrolib/core/io/ini/util.py
```diff
@@ -58,6 +58,40 @@ def split(v: Any):
...
+    *field_names, length_name: str, length_incr: int = 0, id_name: str
+):
+    """
+    Get a root_validator that checks the correct length of several list fields in an object.
+
+    Args:
+        *field_names (str): names of the instance variables that are a list and need checking.
+        length_name (str): name of the instance variable that stores the expected length.
+        length_incr (int): Optional extra increment of length value (e.g., to have +1 extra value in lists).
+        id_name (str): name of the ``id`` instance variable that will be printed in an error message string.
```
priscavdsluis, line 71: This argument should not be necessary with the above suggestion.

hydrolib/core/io/ini/util.py
```diff
@@ -58,6 +58,40 @@ def split(v: Any):
     return validator(*field_name, allow_reuse=True, pre=True)(split)
 
 
+def make_list_length_root_validator(
+    *field_names, length_name: str, length_incr: int = 0, id_name: str
```
priscavdsluis, line 62: It is good practice to put the optional argument at the end of the method signature. 
In this case, you wouldn't need to specify the key `id_name` each time you call the method.

hydrolib/core/io/ini/util.py
```diff
@@ -58,6 +58,40 @@ def split(v: Any):
...
+
+        length = length + length_incr
+
+        for field_name in field_names:
+            field = values.get(field_name)
+            if field is not None and not (
+                isinstance(field, list) and len(field) == length
+            ):
+                raise ValueError(
+                    f"Number of values for {field_name} should be equal to the {length_name} value ({id_name}={values[id_name]})."
```
priscavdsluis, line 88: This message is technically not correct when the `length_incr` is greater than 0.

hydrolib/core/io/ini/models.py
```diff
@@ -25,6 +25,7 @@
 from hydrolib.core.io.ini.util import (
     get_from_subclass_defaults,
```
priscavdsluis, line 26: A general comment:
I think the cross sections should have their own namespace like roughness (friction) now has.

tests/io/test_crosssection.py
```diff
@@ -0,0 +1,267 @@
...
+    CircleCrsDef,
+    CrossDefModel,
+    CrossSectionDefinition,
+    RectangleCrsDef,
+    XYZCrsDef,
+    YZCrsDef,
+    ZWCrsDef,
+    ZWRiverCrsDef,
+)
+from hydrolib.core.io.ini.parser import Parser, ParserConfig
```
priscavdsluis, line 19: I think with the duplicate behaviour within and between the tested classes, you _could_ have introduced some parameterized tests to avoid repetitive tests and to cover all untested behaviour easily. Just a tip!

priscavdsluis (changes requested):
@arthurvd 
Some comments and suggestions.

hydrolib/core/io/ini/models.py
arthurvd: With this code I wanted to catch if the user somehow specifies the empty string for frictionids:
```
frictionid = 
```

hydrolib/core/io/ini/models.py
arthurvd: Cool! I have now added this subroutine to ```CrossSectionDefinition``` in an upcoming commit.
However, I cannot use this inside the root_validator as created by ```def make_list_length_root_validator```, as that one does not have access to the current object (it only receives the static class via ```cls```).

hydrolib/core/io/ini/util.py
arthurvd: I will change it, but in this case: aren't keywords arguments needed anyway in the call? Since they form the separation between the variadic positional arguments and the keywords arguments, right?

hydrolib/core/io/ini/models.py
arthurvd: Yes, I fully agree. And I will do that later in a separate issue, because this issue about unit tests was already work-in-progress --> #149.

hydrolib/core/io/ini/util.py
arthurvd: Fixed in an upcoming commit.

hydrolib/core/io/ini/models.py
```diff
@@ -252,6 +252,9 @@ class Comments(INIBasedModel.Comments):
     type: str = Field(alias="type")
     thalweg: Optional[float]
 
+    def _get_identifier(self, data: dict) -> Optional[str]:
+        return data["id"]
```
priscavdsluis, line 256: If validation fails because the id is missing, this will raise an additional exception because it cannot find the key.
It is safe to use `data.get("id")`

arthurvd: I have fixed all of these in the commit dc72c60 below.

hydrolib/core/io/ini/util.py
priscavdsluis: Yes that is true actually!

priscavdsluis (approved):

PR #148, opened 2021-12-03
Added description of our overrides.

evetion:


docs/topics/pydantic.md
```diff
@@ -0,0 +1,39 @@
...
+We override both `__new__` and `__init__` to cache models from disk. 
+Specifically we also use it to load a filepath from disk.
+
+This mostly happens in `validate`, a Pydantic override, to make
+sure that if we initialize with a single unnamed argument (Pydantic only accepts keyword arguments)
+we try to parse it as a filepath.
+
+### DIMR `__init__` override
+We override the `__init__` method of DIMR to parse any underlying FileModels.
+These are not loaded automatically because they're deeply nested in `components`.
```
BeardedPlatypus, line 21: This has been changed in [the absolute file path issue](https://github.com/Deltares/HYDROLIB-core/blob/6133753375bacc869300c51513adc9c617a866c6/hydrolib/core/io/dimr/models.py#L269), it now does resolve them with a validator, otherwise we have problems with loading the paths correctly.

Not that this affects this review, I will adjust it as part of that issue.

BeardedPlatypus (approved):

Issue #138, opened 2021-11-30
Add RR network topology node file


arthurvd:
Support read-write-store in memory of the contents of RR's network topology **node file**.

Note: this file is related to RR's network topology link file, see issue #140.

Note 2: the file below, and also for related issues all uniquely belong to the RR kernel. Consider grouping these in an ```io/rr/``` subdirectory, instead of a toplevel file extension directory?

**Default filename:** 3b_nod.tp
**Referenced in:** fnm file / ```RainfallRunoffModel.node_data```.
**Format:** ASCII
**Structure:** Optional header line + list of node lines, each node line in file is one record, representing a network node.
**Header syntax:**
The **optional** header line of this file contains in the first positions the string ‘BBB2.2’.
**Node lines syntax:**
The keyword ```NODE``` is used to mark the beginning of a record, and in lower case ```node``` characters it marks the end of a record.
Example line:
```
NODE id ’3B72’ nm ’Node1’ ri ’2’ mt 1 ‘6’ nt 23 ObID ‘SBK_SBK-3B-REACH’ px 11404.2 py 123768.5 node
```
With

- ```id``` node identification ('single quoted character string')
- ```nm``` name of the node ('single quoted character string')
- ```ri``` branch identification ('single quoted integer')
  When not applicable, use '-1'.
- ```mt 1``` model nodetype ('single quoted integer')
  Allowed integer values: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 14, 15, 16, 23. See remark 4 in User Manual below for meaning.
- ```nt``` netter nodetype (integer)
  Must match the model node type using the equivalence table below.
- ```ObID``` Object id ('single quoted character string')
- ```px``` position X (X coordinate) (floating point literal)
- ```py``` position Y (Y coordinate) (floating point literal)
(from: https://content.oss.deltares.nl/delft3d/manuals/SOBEK_User_Manual.pdf#subsection.D.19.13)

Mapping table model node type to netter node type:
(some netter nodetypes map to the same model nodetype. Specifically for ```mt 1 '6'``` model boundaries, the actual Netter type is identified by field ObID.)
|  ```mt 1``` model nodetype | ```nt``` netter nodetype | Node type | ObID |
| ----------- | ----------- |----------- |----------- |
| '1' | 43 | Paved area | '3B_PAVED' |
| '2' | 44 | Unpaved area | '3B_UNPAVED' |
[...]

PR #154, opened 2021-12-06
#138: Add RR network topology node file

priscavdsluis:


tests/test_utils.py
```diff
@@ -35,3 +35,24 @@ def test_given_args_expected_result(self, input_str: str):
...
+class TestGetListIndexSafely:
+    @pytest.mark.parametrize(
+        "list, item, start, end, exp_index",
+        [
+            pytest.param([0, 1, 2], 0, 0, 2, 0),
+            pytest.param([0, 1, 2], 1, 0, 2, 1),
+            pytest.param([0, 1, 2], 2, 0, 2, 2),
+            pytest.param([0, 1, 2], 3, 0, 2, -1),
+            pytest.param([0, 1, 2], 0, 0, 1, 0),
+            pytest.param([0, 1, 2], 1, 0, 1, 1),
```
arthurvd, line 49: If you'd change the behavior of the ```end``` argument (see other review remark), then this test should return -1.

hydrolib/core/io/rr/network/serializer.py
```diff
@@ -0,0 +1,34 @@
...
+        Attributes:
+            path (Path): The path to the destination file.
+            data (Dict): The data to be serialized.
+        """
+
+        path.parent.mkdir(parents=True, exist_ok=True)
+
+        with path.open("wb") as f:
+            for node in data["node"]:
+                line = f"NODE {NodeFileSerializer._to_line(node)} node{os.linesep}"
```
arthurvd, line 20: Not super important, but why not add the "NODE..node" also to the function ```_to_line()```?

hydrolib/core/io/rr/network/parser.py
```diff
@@ -0,0 +1,74 @@
...
+
+        nodes = []
+
+        key_start = "NODE"
+        key_end = "node"
+
+        length_parts = len(parts)
+
+        index = 0
+        while index < length_parts:
```
arthurvd, line 33: Is it not safer for (large) files to replace the above ```file.read()``` and this while loop here by:
```
file = open(path)
# ...
for line in file:
    # ...
```

arthurvd: line-by-line reading is sufficient, because RR requires each node record to be on a single "NODE...node" line.

hydrolib/core/io/rr/network/models.py
```diff
@@ -0,0 +1,107 @@
...
+    54: 16,
+    69: 23,
+}
+
+
+class Node(BaseModel):
+    """Represents a node from the topology node file."""
+
+    id: str = Field(alias="id")
+    name: str = Field(alias="nm")
```
arthurvd, line 36: apologies, I just found out 'nm' is **optional**.

arthurvd (changes requested):

hydrolib/core/io/rr/network/models.py
```diff
@@ -0,0 +1,107 @@
...
+    modelnodetype: int = Field(alias="mt")
+    netternodetype: int = Field(alias="nt")
+    objectid: str = Field(alias="ObID")
+    xposition: float = Field(alias="px")
+    yposition: float = Field(alias="py")
+
+    def _get_identifier(self, data: dict) -> Optional[str]:
+        return data["id"] if "id" in data else data.get("nm")
+
+    def dict(self, *args, **kwargs):
```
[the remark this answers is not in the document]
priscavdsluis: In `FileModel._save()` the following is called: `self._serialize(self.dict())`
The `self.dict()` data is passed to the provided serializer. 
In this case, I don't want to pass the data to the serializer with keys that resemble the class' field names, but rather the aliases, which are the keys that are also parsed by the parser.
So I update the kwargs by saying that I want to serialize this object by using the field aliases as keys.

hydrolib/core/io/rr/network/models.py
priscavdsluis: No problem, adjusted it :)

hydrolib/core/io/rr/network/serializer.py
priscavdsluis: I had been questioning this but eventually decided on the current solution.
I figured that the `_to_line(node)` method would convert a node to the data (in a string).
And this data should then be enclosed between the two keywords. 
But it does not really matter to me as well :)

tests/data/input/rr_network/3B_NOD_format3.TP
```diff
@@ -0,0 +1,16 @@
+NODE id 'unp_AFW_BOM208-P_2275_1' nm 'unp_AFW_BOM208-P_2275_1' ri
```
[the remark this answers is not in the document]
priscavdsluis: Ok, good to know, I assumed that the `node` and `NODE` were leading in terms of when a node starts and ends. Didn't think it was invalid to have a multiple line node.
Will change it!

hydrolib/core/io/rr/network/parser.py
priscavdsluis: Ok, good to know, I assumed that the node and NODE were leading in terms of when a node starts and ends. Didn't think it was invalid to have a multiple line node.
Will change it!

tests/test_utils.py
priscavdsluis: I have removed this function because it was not needed anymore with the parser changes.

hydrolib/core/utils.py
```diff
@@ -45,3 +45,24 @@ def str_is_empty_or_none(str_field: str) -> bool:
...
+        item ([type]): he item to search.
+        start (int): The included index to start searching.
+        end (int): The included index to end searching.
+
+    Returns:
+        int: The index of `item` if found; otherwise, -1.
+    """
+
+    try:
+        return list.index(item, start, end + 1)
```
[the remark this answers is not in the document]
priscavdsluis: I have removed this function because it was not needed anymore with the parser changes.

hydrolib/core/io/rr/network/models.py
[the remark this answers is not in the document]
arthurvd: ok, clear, thanks :-)

arthurvd (approved):

Issue #140, opened 2021-11-30
Add RR topology link file


arthurvd:
Support read-write-store in memory of the contents of RR's network topology **link file**.

Note: this file is related to RR's network topology node file, see issue #138.

Note 2: the file below, and also for related issues all uniquely belong to the RR kernel. Consider grouping these in an ```io/rr/``` subdirectory, instead of a toplevel file extension directory?

**Default filename:** 3b_link.tp
**Referenced in:** fnm file / ```RainfallRunoffModel.branch_data```. :exclamation: Please rename that attribute to ```RainfallRunoffModel.link_data```
**Format:** ASCII
**Structure:** Optional header line + list of link lines, each link line in file is one record, representing a network link.
**Header syntax:**
The **optional** header line of this file contains in the first positions the string ‘BBB2.2’.
**Link lines syntax:**
The keyword ```BRCH``` is used to mark the beginning of a record, and in lower case ```brch``` characters it marks the end of a record.
Example line:
```
BRCH id ’1’ nm ’Tak1’ ri ’-1’ mt 1 ‘0’ bt 11 ObID ‘3B-LINK’ bn ’1’ en ’2’ brch
```
With
- ```id```  link identification ('single quoted character string')
- ```nm```  name of the link ('single quoted character string') **optional**
- ```ri```  branch identification ('single quoted integer')
- ```mt 1``` model link type ('single quoted integer') **optional, default: 0**
  Allowed integer values: TODO
- ```bt```  branch type (integer)
  No validation needed (as opposed to node type validation in #138).
  Currently known types:

|  ```mt 1``` model linktype | ```nt``` netter branchtype | Link type | ObID |
| ----------- | ----------- |----------- |----------- |
| '0' | 17 | RR - Link | '3B_LINK' |

- ```ObID```  Object identification ('single quoted character string')
- ```bn```  identification of begin node (‘from’ node) ('single quoted character string')
- ```en```  identification of end node (‘to’ node) ('single quoted character string')
(from: https://content.oss.deltares.nl/delft3d/manuals/SOBEK_User_Manual.pdf#subsection.D.19.13)

PR #155, opened 2021-12-07
#140: Add RR network topology link file

priscavdsluis:


hydrolib/core/io/rr/topology/serializer.py
```diff
@@ -0,0 +1,70 @@
+import os
+from pathlib import Path
+
+
+class NodeFileSerializer:
```
BeardedPlatypus, line 5: More of a side note than something you really have to adjust now, but you could separate these to two separate files, and then just make regular unbounded functions, instead of grouping them in classes. Though I think in this particular case it makes sense to group them.

hydrolib/core/io/rr/topology/serializer.py
```diff
@@ -0,0 +1,70 @@
...
+
+        path.parent.mkdir(parents=True, exist_ok=True)
+
+        with path.open("wb") as f:
+            for node in data["node"]:
+                line = f"NODE {NodeFileSerializer._to_line(node)} node{os.linesep}"
+                f.write(line.encode("utf8"))
+
+    @staticmethod
+    def _to_line(node: dict) -> str:
```
BeardedPlatypus, line 26: Do we know the sub types of the dict, if the keys are always strings you could write:

```suggestion
    def _to_line(node: Dict[str, Any]) -> str:
```

BeardedPlatypus (approved):
Added some possible suggestion, but overall it looks good, feel free to merge it back.

hydrolib/core/io/rr/topology/serializer.py
priscavdsluis: I'm still trying to development my senses in regards to "python" structuring. But thanks, will keep it in mind!

tests/io/test_fnm.py
```diff
@@ -7,16 +7,32 @@
...
 
 class TestRainFallRunoffModel:
+    def test_load_from_file(self):
+        path = Path(
+            test_input_dir
+            / "e02"
+            / "c11_korte-woerden-1d"
+            / "dimr_model"
+            / "rr"
+            / "Sobek_3b.fnm"
```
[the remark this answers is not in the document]
priscavdsluis: I like this idea :) will create an issue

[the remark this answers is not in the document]
priscavdsluis: #175

PR #167, opened 2021-12-08
Fixed parsing and serialization of Control elements in DIMR.

### LINKED ISSUE #127: DIMR config has invalid control element for a single component model
Unit test `test_from_scratch_docker()` produces an invalid dimr_config.xml file:
```
<?xml version="1.0" encoding="utf-8"?>
<dimrConfig xmlns="http://schemas.deltares.nl/dimr" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://schemas.deltares.nl/dimr http://content.oss.deltares.nl/schemas/dimr-1.3.xsd">
  <documentation>
    <fileVersion>1.3</fileVersion>
    <createdBy>hydrolib-core 0.1.3</createdBy>
    <creationDate>2021-11-01 19:28:06.141447</creationDate>
  </documentation>
  <control/>
  <component name="test">
    <library>dflowfm</library>
    <workingDir>.</workingDir>
    <inputFile>test.mdu</inputFile>
  </component>
</dimrConfig>
```
The empty `<control/>` is wrong. The correct element should be:
```
  <control>
    <start name="test"/>
  <control/>
```

evetion:
Fix #127

### DIFF
--- a/hydrolib/core/io/dimr/models.py
+++ b/hydrolib/core/io/dimr/models.py
@@ -71,6 +71,11 @@ class Component(BaseModel, ABC):
     def _get_identifier(self, data: dict) -> Optional[str]:
         return data.get("name")
 
+    def dict(self, *args, **kwargs):
+        # Exclude the FileModel from any DIMR serialization.
+        kwargs["exclude"] = {"model"}
+        return super().dict(*args, **kwargs)
+
 
 class FMComponent(Component):
     library: Literal["dflowfm"] = "dflowfm"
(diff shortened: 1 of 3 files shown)


hydrolib/core/io/dimr/models.py
```diff
@@ -218,52 +249,43 @@ class Parallel(BaseModel):
...
     coupler: Optional[List[Coupler]] = []
     waitFile: Optional[str]
     global_settings: Optional[GlobalSettings]
 
-    @validator("component", "coupler", pre=True)
+    @validator("component", "coupler", "control", pre=True)
     def validate_component(cls, v):
         return to_list(v)
 
     def dict(self, *args, **kwargs):
```
priscavdsluis, line 285: nice!

priscavdsluis (approved):

priscavdsluis (commented):
I think the changes look good with regards to the XSD.
But do the changes also solve the original request with regards to a non-empty control, and if not, what was the decided action for this issue?

evetion:
> I think the changes look good with regards to the XSD. But do the changes also solve the original request with regards to a non-empty control, and if not, what was the decided action for this issue?

Good point, almost forget the original issue. I'd say it's up to the user right now. So in a tutorial we should document the following (this is from the test):
```python
dimr.component.append(
   FMComponent(name="test", workingDir=".", inputfile=fm.filepath, model=fm)
}
++++    dimr.control.append(Start(name="test"))
```

@arthurvd  What do you think? There's no easy factory method, it really depends on the component(s) added.

hydrolib/core/io/dimr/models.py
arthurvd: @evetion : what is the difference between `core.utils.to_list()` and `core.io.ini.util.make_list_validator()`? Both are used as validators. Should we keep both or not?

arthurvd:
> > I think the changes look good with regards to the XSD. But do the changes also solve the original request with regards to a non-empty control, and if not, what was the decided action for this issue?
> 
> Good point, almost forgot the original issue. I'd say it's up to the user right now. So in a tutorial we should document the following (this is from the test):
> 
> ```python
> dimr.component.append(
>    FMComponent(name="test", workingDir=".", inputfile=fm.filepath, model=fm)
> }
> ++++    dimr.control.append(Start(name="test"))
> ```
> 
> @arthurvd What do you think? There's no easy factory method, it really depends on the component(s) added.

@evetion : I agree, this is fine for now. I have documented this in commit  8b2a59c, please check it out. Afterwards, PR can be merged.

PR #179, opened 2021-12-10
#76: fixed optionality of all pump fields + validators and unit tests.

### LINKED ISSUE #76: Improve validation of optionally required fields in pump structure
Validation of a structure file shows that some optional fields, and some optionally required fields are now validated too strictly for pump structures.
More specifically:
- Field `orientation` is optional.
- **Only** when `numStages` is given and > 0, the following fields must be present: `controlSide`, `startLevelSuctionSide`, `stopLevelSuctionSide`, `startLevelDeliverySide`, `stopLevelDeliverySide`, and each of these last four list values must list length equal to that `numStages`.
- The optionally required field `controlSide` should be an enum that only supports values "suctionSide", "deliverySide" or "both". It has no default.
- **Only** when `numReductionLevels` is given and > 0, the following fields must be present: `head`, `reductionFactor`, and each of these list values must list length equal to that `numReductionLevels`.

arthurvd:
Fixes #76.

### DIFF
--- a/hydrolib/core/io/ini/util.py
+++ b/hydrolib/core/io/ini/util.py
@@ -72,11 +72,11 @@ def make_list_length_root_validator(
         length_name (str): name of the instance variable that stores the expected length.
         length_incr (int): Optional extra increment of length value (e.g., to have +1 extra value in lists).
         list_required_with_length (obj:`bool`, optional): Whether each list *must* be present if the length
-            attribute is present in the input values. Default: False. If False, list length is only checked
-            for the lists that are not None.
+            attribute is present (and > 0) in the input values. Default: False. If False, list length is only
+            checked for the lists that are not None.
     """
 
-    def validate_correct_length(cls, values):
+    def validate_correct_length(cls, values: dict):
         length = values.get(length_name)
         if length is None:
             # length attribute not present, possibly defer validation to a subclass.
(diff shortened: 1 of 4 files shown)


hydrolib/core/io/ini/util.py
```diff
@@ -81,12 +87,16 @@ def validate_correct_length(cls, values):
 
         for field_name in field_names:
             field = values.get(field_name)
-            if field is not None and not (
-                isinstance(field, list) and len(field) == length
-            ):
+            if field is not None:
+                if not (isinstance(field, list) and len(field) == length):
```
priscavdsluis, line 91: I know this check was already there, but now that I think about it, the `if not (isinstance(field, list)` should not really be necessary right? If someone adds fields to this function that are not lists, it is a programming mistake not a valid situation. Or am I missing something? If it is an invalid function input I would remove this first check.

hydrolib/core/io/structure/models.py
```diff
@@ -395,6 +409,23 @@ class Pump(Structure):
...
+    _orientation_validator = get_enum_validator("orientation", enum=Orientation)
+
+    _check_list_length = make_list_length_root_validator(
+        "startlevelsuctionside",
+        "stoplevelsuctionside",
+        "startleveldeliveryside",
+        "stopleveldeliveryside",
+        length_name="numstages",
+        list_required_with_length=True,
+    )
```
priscavdsluis, line 421: Please correct me if I am wrong but I understand from the issue that if e.g. `numstages == 0` than all those fields are allowed to be `None` . But I don't see where that is handled.

priscavdsluis (commented):
@arthur I have two questions actually. Changes look good :)

hydrolib/core/io/ini/util.py
arthurvd: fair point, changed.

hydrolib/core/io/structure/models.py
arthurvd: I've changed it into `elif list_required_with_length and length > 0:`

PR #180, opened 2021-12-10
Add Version checks to all relevant models.

### LINKED ISSUE #94: Check fileVersion numbers and define which version(s) we support
- for each filetype, we should maintain what is the current fileVersion that HYDROLIB-core supports.
- As such, after parsing a file, a check should be done on the file's fileVersion and the HYDROLIB-core supported fileVersion.
- fileVersion numbers are now stored as a float, but note that we want to support this:
  2.1 --> major=2, minor=1
  2.10 --> major=2, minor=10
- Suggestion for what forward/backward compatibility HYDROLIB-core should support:
  - file major < HYDROLIB major --> not supported
  - file major == HYDROLIB major and file minor <= HYDROLIB minor --> supported
  - file major == HYDROLIB major and file minor > HYDROLIB minor --> not supported
  - file major > HYDROLIB major --> not supported
- A clean error message should appear in case of non-compatibility (something like "File too new/old for HYDROLIB-core readers."

evetion:
And lowered bounds of default versions.
Fixes #94.

### DIFF
--- a/hydrolib/core/io/mdu/models.py
+++ b/hydrolib/core/io/mdu/models.py
@@ -2,7 +2,6 @@ from pathlib import Path
 from typing import Dict, List, Literal, Optional, Union
 
 from pydantic import Field
-from hydrolib.core.utils import FMVersion
 
 from hydrolib.core.io.crosssection.models import CrossDefModel, CrossLocModel
 from hydrolib.core.io.ext.models import ExtModel
@@ -13,6 +12,7 @@ from hydrolib.core.io.net.models import NetworkModel
 from hydrolib.core.io.polyfile.models import PolyFile
 from hydrolib.core.io.structure.models import StructureModel
 from hydrolib.core.io.xyz.models import XYZModel
+from hydrolib.core.utils import FMVersion
 
 
 class General(INIGeneral):
--- a/hydrolib/core/io/mdu/models.py
+++ b/hydrolib/core/io/mdu/models.py
@@ -2,6 +2,7 @@ from pathlib import Path
 from typing import Dict, List, Literal, Optional, Union
 
 from pydantic import Field
+from hydrolib.core.utils import FMVersion
 
 from hydrolib.core.io.crosssection.models import CrossDefModel, CrossLocModel
 from hydrolib.core.io.ext.models import ExtModel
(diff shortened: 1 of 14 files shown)


evetion:
@arthurvd Downside to this work is that we now write the lower bound version number to files.

hydrolib/core/utils.py
```diff
@@ -69,3 +72,55 @@ def get_substring_between(source: str, start: str, end: str) -> Optional[str]:
...
+
+
+class HVersion(SemVer):
+    @classmethod
+    def coerce(cls, *args, **kwargs):
+        version = super().coerce(*args, **kwargs)
+        return cls(**version.__dict__)
+
+
+class FMVersion(HVersion):
```
arthurvd, line 84: I don't quite like this FM-related class here, inside a "neutral" utils namespace.
Also it suggest that this is "the" FM version, whereas it is only the numbering schema, so consider renaming this class to: `class FMStyleVersion`, or `DHydroStyleVersion`, or `ZeroPrefixedVersion`.

hydrolib/core/io/mdu/models.py
```diff
@@ -12,14 +12,15 @@
...
+from hydrolib.core.utils import FMVersion
 
 
 class General(INIGeneral):
     _header: Literal["General"] = "General"
     program: str = Field("D-Flow FM", alias="program")
     version: str = Field("1.2.94.66079M", alias="version")
     filetype: Literal["modelDef"] = Field("modelDef", alias="fileType")
-    fileversion: str = Field("1.09", alias="fileVersion")
+    fileversion: FMVersion = Field(FMVersion("1.9.0"), alias="fileVersion")
```
arthurvd, line 23: I understand the attempt, yet I find it reads quite confusing.

arthurvd (commented):
Hey @evetion , I've added some comments, but I am in doubt about whether we should squeeze this into the 0.1.6 release, as I think it deserves some further thinking.
Writing "1.09" FM-style version as `FMVersion("1.9.0") is pretty confusing. Maybe we should consider constructors using major, minor as indifvidual arguments.

Anyways, food for thought.

evetion:
> but I am in doubt about whether we should squeeze this into the 0.1.6 release, as I think it deserves some further thinking.

Agreed, while most of the comments can be easily resolved, it's most about the lower/upper bound discussion (what do we write as version) that needs some thought.

PR #181, opened 2021-12-10
#98: introduced enum for FrictionType everywhere.

### LINKED ISSUE #98: Make enum for frictionType
Similar to allowedFlowDir, we need also an enum for friction types.
- allowed values in input: Chezy, Manning, wallLawNikuradse, WhiteColebrook, StricklerNikuradse, Strickler, deBosBijkerk
- to be used in structure models for: Bridge, Culvert, in ini models for (Friction) Global
- in the future also detailed cross section Definition objects

arthurvd:
Fixes #98.


tests/data/input/dflowfm_individual_files/crsdef1.ini
```diff
@@ -15,7 +15,7 @@
     type                  = circle
     thalweg               = 0.0
     diameter              = 2.0
-    frictionType          = 3
```
evetion, line 18: Is this actually a IntEnum?

hydrolib/core/io/friction/models.py
```diff
@@ -1,18 +1,38 @@
...
+
+    chezy = "Chezy"  # Chézy C [m 1/2 /s]
+    manning = "Manning"  # Manning n [s/m 1/3 ]
+    walllawnikuradse = "wallLawNikuradse"  # Nikuradse k n [m]
+    whitecolebrook = "WhiteColebrook"  # Nikuradse k n [m]
+    stricklernikuradse = "StricklerNikuradse"  # Nikuradse k n [m]
+    strickler = "Strickler"  # Strickler k s [m 1/3 /s]
+    debosbijkerk = "deBosBijkerk"  # De Bos-Bijkerk γ [1/s]
+
+    allowedvaluestext = "Possible values: Chezy, Manning, wallLawNikuradse, WhiteColebrook, StricklerNikuradse, Strickler, deBosBijkerk."
```
evetion, line 33: This is unused. Other `Enum`s tend to use this in the Comment default for this field. I feel it's a bit ugly, as it now becomes a possible option.

hydrolib/core/io/friction/models.py
```diff
@@ -1,18 +1,38 @@
...
 logger = logging.getLogger(__name__)
 
 
+class FrictionType(str, Enum):
+    """
+    Enum class containing the valid values for the frictionType
+    attribute in several subclasses of Structure/CrossSection/friction.models.
+    """
+
+    chezy = "Chezy"  # Chézy C [m 1/2 /s]
```
evetion, line 25: These comments might be better placed in the docstring?

evetion (changes requested):
I'm unsure about serialization, I don't think that's tested in this PR yet. You might want to take a look at the `use_enum_values` option in the class Config.

tests/data/input/dflowfm_individual_files/crsdef1.ini
arthurvd: No, it's not. In user input this is definitely a string enum. Only later in kernel it becomes an int enum, but that is not relevant for us here.

hydrolib/core/io/friction/models.py
arthurvd: I agree, so removed it. In particular because of your last "option" point.

hydrolib/core/io/friction/models.py
arthurvd: Done, as postfix, now nice apidocs generated.

arthurvd:
> I'm unsure about serialization, I don't think that's tested in this PR yet. You might want to take a look at the `use_enum_values` option in the class Config.

I have added `test_save()`, which confirms that this works well. The `use_enum_values` is True in our `BaseModel` already.

evetion (approved):

Issue #75, opened 2021-10-19
Improve validation of optionally required fields in culvert structure


arthurvd:
Validation of a structure file shows that some optional fields, and some optionally required fields are now validated too strictly for culvert structures.
More specifically:
- **Only** when `valveOnOff` equals 1/true, the following fields must be present: `valveOpeningHeight`, `numLossCoeff`, `relOpening`, `lossCoeff`. Otherwise all these latter four fields may be omitted.
- Field `subType` is optional and has default value `culvert` when missing in input.
- **Only** when `subType` equals "culvert", the field `bendLossCoeff` must be present. Otherwise not.

PR #182, opened 2021-12-13
Feature/119 inifield file

### LINKED ISSUE #119: Support initial field files
https://content.oss.deltares.nl/delft3d/manuals/D-Flow_FM_User_Manual_1D2D.pdf#section.D.2

arthurvd:
Fixes #119. Both the `iniFieldFile`, as well as 1D field INI file format (see UM sec. D.2 and D.4.1).


hydrolib/core/io/ini/util.py
```diff
@@ -82,15 +85,16 @@ def validate_correct_length(cls, values: dict):
             # length attribute not present, possibly defer validation to a subclass.
             return values
 
-        length = length + length_incr
+        length = max(length + length_incr, min_length)
```
priscavdsluis, line 88: if `min_length` here is the greatest value, the code says that the lists must exactly have this number of values, but cannot have more than that. Is that correct? The key word says it is the minimal required length suggesting that the number of values is allowerd to be higher.

hydrolib/core/io/inifield/models.py
```diff
@@ -0,0 +1,244 @@
...
+    constant = "constant"  # only with dataFileType=polygon .
+    triangulation = "triangulation"  # Delaunay triangulation+linear interpolation.
+    averaging = "averaging"  # grid cell averaging.
+
+    allowedvaluestext = "Possible values: constant, triangulation, averaging."
+
+
+class Operand(str, Enum):
+    """
+    Enum class containing the valid values for the operand
```
priscavdsluis, line 51:
```suggestion
    Enum class containing the valid values for the interpolationMethod
```

hydrolib/core/io/inifield/models.py
```diff
@@ -0,0 +1,244 @@
...
+        value: Optional[str] = Field(
+            "Only for dataFileType=polygon. The constant value to be set inside for all model points inside the polygon."
+        )
+
+    comments: Comments = Comments()
+
+    quantity: str = Field(alias="quantity")
+    datafile: Path = Field(alias="dataFile")
+    datafiletype: DataFileType = Field(alias="dataFileType")
+    interpolationmethod: Optional[InterpolationMethod] = Field(
```
priscavdsluis, line 165: Should interpolationMethod really be optional? I do not see it in the manual.

hydrolib/core/io/inifield/models.py
```diff
@@ -0,0 +1,244 @@
...
+    operand: Optional[Operand] = Field(Operand.override, alias="operand")
+    averagingtype: Optional[AveragingType] = Field(
+        AveragingType.mean, alias="averagingType"
+    )
+    averagingrelsize: Optional[NonNegativeFloat] = Field(1.01, alias="averagingRelSize")
+    averagingnummin: Optional[PositiveInt] = Field(1, alias="averagingNumMin")
+    averagingpercentile: Optional[NonNegativeFloat] = Field(
+        0, alias="averagingPercentile"
+    )
+    extrapolationmethod: Optional[bool] = Field(False, alias="extrapolationMethod")
```
priscavdsluis, line 177: I do not think this will go correctly without validator unfortunately, because possible values for `extrapolationMethod` are "yes" and "no", instead of "true" and "false" or "0" and "1"

hydrolib/core/io/inifield/models.py
```diff
@@ -0,0 +1,244 @@
...
+    averagingtype_validator = get_enum_validator("averagingtype", enum=AveragingType)
+    locationtype_validator = get_enum_validator("locationtype", enum=LocationType)
+
+    @validator("value", always=True)
+    @classmethod
+    def _validate_value_and_filetype(cls, v, values: dict):
+        if v is not None and values.get("datafiletype") != DataFileType.polygon:
+            raise ValueError(
+                f"When value={v} is given, dataFileType={DataFileType.polygon} is required."
+            )
```
priscavdsluis, line 195: Shouldnt it be the other way around?
When datafiletype == DataFileType.polygon, `value` should not be None. (that is how I interpret the manual)

hydrolib/core/io/ini/util.py
priscavdsluis: Also SC is complaining about the cognitive complexity of this method. Mayme refactor a little by extracting the lower body of this function.

tests/io/test_inifield.py
```diff
@@ -0,0 +1,165 @@
...
+    LocationType,
+    Operand,
+    ParameterField,
+)
+
+from ..utils import WrapperTest, invalid_test_data_dir, test_data_dir
+
+
+class TestIniField:
+    def _get_datafiletype_cases() -> List:
```
priscavdsluis, line 27: There are a couple of SC bugs regarding these functions. Please add `self` to the function signature.

tests/io/test_inifield.py
```diff
@@ -0,0 +1,165 @@
...
+
+        return inifield_values
+
+    @pytest.mark.parametrize(
+        "attribute,input,expected",
+        _get_datafiletype_cases()
+        + _get_interpolationmethod_cases()
+        + _get_operand_cases()
+        + _get_averagingtype_cases()
+        + _get_locationtype_cases(),
```
priscavdsluis, line 83: Cool, I did not know this was possible :)

hydrolib/core/io/onedfield/models.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def _get_identifier(self, data: dict) -> Optional[str]:
+        return data.get("branchid")
+
+
+class OneDFieldModel(INIModel):
+    """
+    The overall 1D field model that contains the contents of a 1D field file.
+
+    This model is typically referenced under a [FMModel][hydrolib.core.io.mdu.models.FMModel]`.geometry.inifieldfile[..].initial[..].datafile`.
```
priscavdsluis, line 112: This does not exist yet; `OneDFieldModel` is never used. currently.

priscavdsluis (changes requested):
Left a couple of comments/questions. Also check the SonarCloud status.

hydrolib/core/io/inifield/models.py
arthurvd: actually correct here, but you must have seen that it was wrong in class `InterpolationMethod`. I've fixed that.

hydrolib/core/io/inifield/models.py
arthurvd: We should use both:
* when v is given, dataFileType cannot be != polygon (my present check),
  or in other words, when dataFileType != polygon then v should not be given.
* And also, when dataFileType==polygon, then v must be given. I will add that once #75 is merged to main.

hydrolib/core/io/onedfield/models.py
arthurvd: I have corrected this in this documentation, and added a new refinement issue for this: #184.

hydrolib/core/io/inifield/models.py
priscavdsluis: Oops sorry!

tests/io/test_inifield.py
arthurvd: Solved by making static variables, because the functions as true object functions could not be called in `@pytest.mark.parametrize`. I also could not get `@staticmethod` to work, therefore, as static class variables. Now alle these tests work well.

hydrolib/core/io/inifield/models.py
arthurvd: It is optional, because when `dataFileType=1dField`, no interpolation method is necessary

hydrolib/core/io/inifield/models.py
arthurvd: I just found out a bug in the kernel here, so I will have that fixed first, and have opened a followup issue here: #186.

arthurvd:
Code smell now also solved.

hydrolib/core/io/ini/util.py
arthurvd: I corrected the documentation/explanation + fixed code smell.

hydrolib/core/io/inifield/models.py
```diff
@@ -186,6 +186,12 @@ class Comments(INIBasedModel.Comments):
     averagingtype_validator = get_enum_validator("averagingtype", enum=AveragingType)
     locationtype_validator = get_enum_validator("locationtype", enum=LocationType)
 
+    # TODO: Enable once PR #75 has been merged to main:
```
priscavdsluis, line 189: You can enable this now

hydrolib/core/io/ini/util.py
```diff
@@ -76,28 +77,55 @@ def make_list_length_root_validator(
         list_required_with_length (obj:`bool`, optional): Whether each list *must* be present if the length
             attribute is present (and > 0) in the input values. Default: False. If False, list length is only
             checked for the lists that are not None.
+        min_length (int): minimum for list length value, overrides length_name value if that is smaller.
+            For example, to require list length 1 when length value is given as 0.
     """
 
+    def _get_incorrect_length_validation_message() -> str:
+        """Make a string with a validation message, ready to be format()ed with
```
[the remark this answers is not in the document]
arthurvd: This was actually intentional ;-)

hydrolib/core/io/inifield/models.py
arthurvd: Done, see last commit.

PR #206, opened 2022-01-29
#205: Allow empty friction specification in all crossdef types

### LINKED ISSUE #205: CrossDefModel validation should not require any friction specification
**Describe the bug**
Initially reported by @LisaWeijers in #199. When reading a cross section definition that has not any friction specification, the validation currently fails. This is wrong, because D-HYDRO always has default values for the `frictionId`(`s`).

**To Reproduce**
1. Take example file [crsdef.ini.txt](https://github.com/Deltares/HYDROLIB-core/files/7854043/crsdef.ini.txt) (rename to .ini)
2. Load the file using:
```python
from hydrolib.core.io.crosssection.models import CrossDefModel
crossdef_new = CrossDefModel(modeldir / "crsdef.ini")
```
3. See error
```
Exception has occurred: ValidationError
796 validation errors for CrossDefModel
crsdef.ini -> definition -> 1389 -> KBR-1468 -> __root__
  Cross section is missing any friction specification. (type=value_error)
```

**Expected behavior**
Validation error should not occur, empty frictionId/Type/Value is allowed.

 - OS: Windows (N/A)
 - Version: 0.2.0

arthurvd:
Fixes #205

### DIFF
--- a/hydrolib/core/io/crosssection/models.py
+++ b/hydrolib/core/io/crosssection/models.py
@@ -137,13 +137,7 @@ class CrossSectionDefinition(INIBasedModel):
             frictiontype = values.get(frictiontype_attr) or ""
             frictionvalue = values.get(frictionvalue_attr) or ""
 
-            if frictionid == "":
-                if frictiontype == "" or frictionvalue == "":
-                    # No validation error: frictionid will always default to some
-                    # value (Main/FloodPlain1/FloodPlain2), but will not fill that
-                    # default here, to keep output blank upon serialization.
-                    pass
-            else:
+            if frictionid != "":
                 if frictiontype != "" or frictionvalue != "":
                     raise ValueError(
                         f"Cross section has duplicate friction specification (both {frictionid_attr} and {frictiontype_attr}/{frictionvalue_attr})."
--- a/hydrolib/core/io/crosssection/models.py
+++ b/hydrolib/core/io/crosssection/models.py
@@ -139,9 +139,10 @@ class CrossSectionDefinition(INIBasedModel):
 
             if frictionid == "":
                 if frictiontype == "" or frictionvalue == "":
-                    raise ValueError(
-                        f"Cross section is missing any friction specification."
-                    )
+                    # No validation error: frictionid will always default to some
+                    # value (Main/FloodPlain1/FloodPlain2), but will not fill that
+                    # default here, to keep output blank upon serialization.
+                    pass
             else:
                 if frictiontype != "" or frictionvalue != "":
                     raise ValueError(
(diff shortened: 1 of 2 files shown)


LisaWeijers:
Works like a charm! :) 

@arthurvd do you want me to review/approve this then? I tried it, and it works, but I wouldn't dare to say if this is a good/robust solution that fits Hydrolib-core's code standards etc

hydrolib/core/io/crosssection/models.py
```diff
@@ -139,9 +139,10 @@ def validate_friction_specification(cls, values):
 
             if frictionid == "":
                 if frictiontype == "" or frictionvalue == "":
-                    raise ValueError(
-                        f"Cross section is missing any friction specification."
-                    )
+                    # No validation error: frictionid will always default to some
```
priscavdsluis, line 142: So if the kernel is not provided a friction it calculate with the default?

tests/io/test_crosssection.py
```diff
@@ -157,16 +156,16 @@ def test_create_a_zwrivercrsdef_with_wrong_list_lengths(self):
...
+        crsdef = ZWRiverCrsDef(
+            id=csdefid,
+            numlevels=2,
+            levels=[-2, 3],
+            flowwidths=[11, 44],
+        )
+        assert crsdef.id == csdefid
+        assert crsdef.numlevels == 2
+        assert crsdef.levels == [-2, 3]
+        assert crsdef.flowwidths == [11, 44]
```
priscavdsluis, line 168: Since validation passes, maybe also test for the other properties? 
Also, since the missing of the friction is now not a "special" case anymore (no if-statement to check for this situation), you normally don't test for this. But since we already have a test in place, I would leave it :)

priscavdsluis (changes requested):
See comments.

hydrolib/core/io/crosssection/models.py
arthurvd: indeed, the frictionId "Main", etc., by the kernel. Intentionally leaving it blank on the Python / serialization side.

hydrolib/core/io/crosssection/models.py
```diff
@@ -139,9 +139,10 @@ def validate_friction_specification(cls, values):
...
+                    # value (Main/FloodPlain1/FloodPlain2), but will not fill that
```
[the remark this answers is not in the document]
arthurvd: ok done, removed the comment as well (pasted in commit message for traceability)

tests/io/test_crosssection.py
arthurvd: I've added testing on friction list fields to be None (because that's what this test is about.
all the other optional fields I've kept out of this test.

PR #215, opened 2022-02-17
ignore location fields for compound type structures

### LINKED ISSUE #214: Compound structures are written with inherited but irrelevant location fields
@LisaWeijers wrote in #211:
Error messages in DHydro

A few things that I notice here:
- The compound structure that is included in structure_temp.ini is exported with an empty fields (third message in screenshot). It's about the line I marked in the screenshot below
Compound structure snippet

_Originally posted by @LisaWeijers in https://github.com/Deltares/HYDROLIB-core/issues/211#issuecomment-1029966848_

**Cause:**
Class `Compound` inherits from `Structure`, but none of the location fields are relevant/allowed in a compound.

From the User Manual:
* _compound structure should **not** contain a location specification. This will be retrieved from the individual structures._
* _A compound structure **does** require its own id and name fields, because the output in the history file will also contain total output for the compound’s underlying structures together and this will appear under its own compound id._

**Related question:**
In the above screenshot, left file, why are the num/x/ycoordinate fields written as empty, but the other location fields (branchid + chainage) are *not at all* included?
Answer: this is cause by this code:
```python
    def _exclude_fields(self) -> Set:
        # exclude the unset props like coordinates or branches
        if self.branchid is not None:
            exclude_set = {"numcoordinates", "xcoordinates", "ycoordinates"}
        else:
            exclude_set = {"branchid", "chainage"}
```
For 'normal' structures (non-compound), this is good code.

**Possible solution:**
do not make Compound a subclass of Structure, but then make sure it is still allowed in a `Structure` model file.

arthurvd:
Fixes #214.
* serialization now hides all location fields, thanks to improved _exclude_fields
* I've prepared improved validation upon parsing, but left the code commented out, because the D-HYDRO Suite 1D2D currently still writes branchId+chainage for compounds (but this will have to change)
* also added extra unit test for the compound reading.

### DIFF
--- a/hydrolib/core/io/structure/models.py
+++ b/hydrolib/core/io/structure/models.py
@@ -68,6 +68,10 @@ class Structure(INIBasedModel):
     xcoordinates: Optional[List[float]] = Field(None, alias="xCoordinates")
     ycoordinates: Optional[List[float]] = Field(None, alias="yCoordinates")
 
+    _loc_coord_fields = {"numcoordinates", "xcoordinates", "ycoordinates"}
+    _loc_branch_fields = {"branchid", "chainage"}
+    _loc_all_fields = _loc_coord_fields | _loc_branch_fields
+
     _split_to_list = get_split_string_on_delimiter_validator(
         "xcoordinates", "ycoordinates"
     )
@@ -97,8 +101,18 @@ class Structure(INIBasedModel):
         filtered_values = {k: v for k, v in values.items() if v is not None}
         structype = filtered_values.get("type", "").lower()
 
+        if structype == "compound" or issubclass(cls, Compound):
+            loc_in_model = any(
+                k.lower() in cls._loc_all_fields for k in filtered_values.keys()
+            )
+            # TODO: issue 214: enable the assert below once the D-HYDRO Suite 1D2D
+            # has fixed issue FM1D2D-1935 for compound structures.
+            # assert (
+            #     not loc_in_model
+            # ), f"No `num/x/yCoordinates` nor `branchId+chainage` are allowed for a {structype} structure."
+
         # TODO This seems to be a bit of a hack.
-        if not (structype == "compound" or issubclass(cls, (Compound, Dambreak))):
+        elif not (structype == "compound" or issubclass(cls, (Compound, Dambreak))):
             # Compound structure does not require a location specification.
             only_coordinates_structures = dict(
                 longculvert="LongCulvert", dambreak="Dambreak"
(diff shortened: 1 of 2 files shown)


hydrolib/core/io/structure/models.py
```diff
@@ -214,11 +228,13 @@ def validate(cls, v):
         return super().validate(v)
 
     def _exclude_fields(self) -> Set:
-        # exclude the unset props like coordinates or branches
-        if self.branchid is not None:
-            exclude_set = {"numcoordinates", "xcoordinates", "ycoordinates"}
+        # exclude the non-applicable, or unset props like coordinates or branches
+        if self.type is "compound":
+            exclude_set = self._loc_all_fields
```
evetion, line 233: More readable to use `_loc_coord_fields | _loc_branch_fields` here?

hydrolib/core/io/structure/models.py
```diff
@@ -97,8 +101,18 @@ def check_location(cls, values: dict) -> dict:
         filtered_values = {k: v for k, v in values.items() if v is not None}
         structype = filtered_values.get("type", "").lower()
 
+        if structype == "compound" or issubclass(cls, Compound):
```
evetion, line 104: Maybe leave out these changes completely and put it in another draft PR? It's all TODO without a clearly defined use/release/timeline.

tests/io/test_structure.py
```diff
@@ -205,7 +205,12 @@ def test_create_structure_without_type():
     ],
 )
 def test_parses_type_case_insensitive(input, expected):
-    structure = Structure(type=input, branchid="branchid", chainage="1")
+    if input.lower() != "compound":
```
evetion, line 208: Introducing such logic in a test seems wrong. Better to write another test or rewrite it with fixtures? What is it that you're testing?

evetion (changes requested):

hydrolib/core/io/structure/models.py
arthurvd: I prefer to keep it like this: it reflects that a compound never has ay coordinate fields. And it is also consistent with the check_location code on line 106.

tests/io/test_structure.py
arthurvd: See commit below for improvement.

hydrolib/core/io/structure/models.py
arthurvd: Done, removed for now. The underlying D3DFMIQ issue is being picked up soon, after which I will decide on what to do with this code fragment.

PR #218, opened 2022-03-10
217: extend to_key such that leading digits are removed from aliased fields.

### LINKED ISSUE #217: Repeated MDU save causes duplicate comment list for 1d2DLinkFile field
Reported by @LisaWeijers in #211:
Alright, it took a while to get back to this because I'm both trying to test if this is robust & I'm working for the profile optimizer at the same time. But I think I figured some things out about some thing

I used this version:

- lists of numbers are saved with wrong seperators
--> probably fixed in PR https://github.com/Deltares/HYDROLIB-core/pull/212.

This works better! I can save a copy of an existing FM model, and the copy can be read by Hydrolib again. So partially solved. 

My intention (which is in hindsight, not the best approach, I found an neater alternative) was to do this:
BASE MODEL --> change a profile --> ITERATION 1 --> change a profile --> ITERATION 2 --> change a profile --> etc
This produced a new error. Here is a base code example that produces the bug:
```
old_mdu_path = r'c:\dhydro\20211102_Developers_tutorial_Hydrolib\code\data\model\moergestels_broek.mdu'
model_folder = Path("playground")

old_mdu = FMModel(Path(old_mdu_path))
mdu_1_path = Path(model_folder)/'test_1' / 'save_1.mdu'
old_mdu.save(mdu_1_path, recurse=True)

mdu_1 = FMModel(mdu_1_path)
mdu_2_path = Path(model_folder)/'test_2' / 'save_2.mdu'
mdu_1.save(mdu_2_path, recurse=True)

mdu_2 = FMModel(mdu_2_path)
print(mdu_2.show_tree())
```
This crashes at `mdu_1.save(mdu_2_path, recurse=True)` with the following output:

```
Traceback (most recent call last):
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\basemodel.py", line 484, in file_load_context
    yield file_loading_context
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\basemodel.py", line 670, in save
    self._save_tree(context)
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\basemodel.py", line 715, in _save_tree
    save_traverser.traverse(self, context)
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\basemodel.py", line 240, in traverse
    acc = self._post_traverse_func(model, acc)  # type: ignore[arg-type]
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\basemodel.py", line 706, in save_post
    model._save()
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\basemodel.py", line 723, in _save
    self._serialize(self.dict())
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\io\ini\models.py", line 190, in _serialize
    write_ini(self._resolved_filepath, self._to_document(), config=config)
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\io\ini\models.py", line 184, in _to_document
    sections.append(value._to_section())
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\io\ini\models.py", line 121, in _to_section
    prop = Property(
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\basemodel.py", line 52, in __init__
    raise e
  File "C:\ProgramData\Anaconda3\envs\hydrolib_resave\lib\site-packages\hydrolib\core\basemodel.py", line 48, in __init__
    super().__init__(**data)
  File "pydantic\main.py", line 406, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for Property
comment
  str type expected (type=type_error.str)
```
What it's trying to work with is this:
`{'key': '1d2dLinkFile', 'value': '', 'comment': ['File containing the custom parameterization of 1D-2D links.', 'File containing the custom parameterization of 1D-2D links.']}`
The value for comment is a list, which shouldn't happen, it should be a string. 

I can work around it for the profile optimizer, namely by always doing my changes on a base model, not on the previous iterations. But hopefully this report can help make hydrolib more robust ;) 

_Originally posted by @LisaWeijers in https://github.com/Deltares/HYDROLIB-core/issues/211#issuecomment-1048903690_

arthurvd:
Fixes #217.

### DIFF
--- a/hydrolib/core/utils.py
+++ b/hydrolib/core/utils.py
@@ -16,6 +16,37 @@ def example(a: float, b: float = 1.0) -> float:
 
 
 def to_key(string: str) -> str:
+    """
+    Construct a key name from a given field name.
+    The given field name may be a Pydantic Field alias, and the key name
+    is intended to be used as a BaseModel class member variable name.
+
+    Args:
+        string (str): input field name
+
+    """
+    # First replace any leading digits, because those are undesirable
+    # in variable names.
+    digitdict = {
+        "0": "zero",
+        "1": "one",
+        "2": "two",
+        "3": "three",
+        "4": "four",
+        "5": "five",
+        "6": "six",
+        "7": "seven",
+        "8": "eight",
+        "9": "nine",
+    }
+    m = re.search(r"^\d+", string)
+    if m:
+        digitstring = string[0 : m.end()]
+        for key, val in digitdict.items():
+            digitstring = digitstring.replace(key, val)
+        string = digitstring + string[m.end() :]
+
+    # Next, replace spaces and hyphens in the potential variable name.
     return string.lower().replace(" ", "_").replace("-", "")
 
 
--- a/hydrolib/core/utils.py
+++ b/hydrolib/core/utils.py
@@ -1,6 +1,6 @@
 from operator import eq, ge, gt, le, lt, ne
 from typing import Any, Callable, List, Optional
-
+import re
 
 def example(a: float, b: float = 1.0) -> float:
     """[summary]
--- a/hydrolib/core/utils.py
+++ b/hydrolib/core/utils.py
@@ -1,6 +1,7 @@
+import re
 from operator import eq, ge, gt, le, lt, ne
 from typing import Any, Callable, List, Optional
-import re
+
 
 def example(a: float, b: float = 1.0) -> float:
     """[summary]
--- a/hydrolib/core/io/mdu/models.py
+++ b/hydrolib/core/io/mdu/models.py
@@ -457,7 +457,7 @@ class Geometry(INIBasedModel):
     crossdeffile: Optional[CrossDefModel] = Field(None, alias="crossDefFile")
     crosslocfile: Optional[CrossLocModel] = Field(None, alias="crossLocFile")
     storagenodefile: Optional[StorageNodeModel] = Field(None, alias="storageNodeFile")
-    onedtwodlinkfile: Optional[Path] = Field(None, alias="1d2dLinkFile")
+    oned2dlinkfile: Optional[Path] = Field(None, alias="1d2dLinkFile")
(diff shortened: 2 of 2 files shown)


hydrolib/core/utils.py
```diff
@@ -16,6 +17,37 @@ def example(a: float, b: float = 1.0) -> float:
...
+    """
+    Construct a key name from a given field name.
+    The given field name may be a Pydantic Field alias, and the key name
+    is intended to be used as a BaseModel class member variable name.
+
+    Args:
+        string (str): input field name
+
+    """
+    # First replace any leading digits, because those are undesirable
```
evetion, line 29: Maybe replace all digits? Bit inconsistent to have oned2dfield? Would simplify the code underneath as well, removing all `re` parts. Or the more extreme and ugly alternative in the other direction, just replace 1 if it occurs, this probably only happens on the `1d2dlinkfile` field and no other fields contain a single `1`?

arthurvd: I did this originally, but changed plans, because a lot of class members then changed, such as:
`nonlin1d`, `teta0`, `uniffrictcoef1d`, etc., etc.
The trick is dirty enough by itself and only intended to avoid invalid Python variable names starting with digits. Beyond that, I don't want to fiddle to much with our model class fields.

PR #222, opened 2022-03-14
Use ModelField attributes to get the split list delimiter

evetion:
From the docs over at https://pydantic-docs.helpmanual.io/usage/validators/#pre-and-per-item-validators

```
you can also add any subset of the following arguments to the signature (the names must match):

    values: a dict containing the name-to-value mapping of any previously-validated fields
    config: the model config
    field: the field being validated. Type of object is pydantic.fields.ModelField.
    **kwargs: if provided, this will include the arguments above not explicitly listed in the signature
```


hydrolib/core/io/ini/util.py
```diff
@@ -25,9 +26,9 @@ def get_split_string_on_delimiter_validator(*field_name: str, delimiter: str = N
         the validator which splits strings on the provided delimiter.
     """
 
-    def split(v: Any):
+    def split(v: Any, field: ModelField):
         if isinstance(v, str):
-            v = v.split(delimiter)
+            v = v.split(field.field_info.extra.get("delimiter", delimiter))
```
arthurvd, line 31: This works well, when a Field explicitly has this attribute.
I tested this for class core.io.mdu.models.Geometry.frictfile, and indeed all tests under `test_basemodel.py/TestFileModel.test_save_and_load_maintains_correct_paths()` stay green.

However, when a field has no own delimiter, but relies on the its class'es default delimiter, then that class delimiter cannot be accessed by the validator (class is unknown/not passed), so parsing list values may be incorrect.

arthurvd (changes requested):
See comment about class level delimiters.

hydrolib/core/io/ini/util.py
evetion: You can also access the class here, but the question is more where to use/store this info. We now do it three times:
- A default for a class (" ")
- A delimiter passed in the creation of the validation function
- A delimiter set on the ModelField

In the above function the default for the class is not used, since all the `get_split_string_on_delimiter_validator` calls specify one.

evetion: Example for access to the class delimiter (but this ignores the delimiter passed with `get_split_string_on_delimiter_validator`)

```python
    def split(cls, v: Any, field: ModelField):
        if isinstance(v, str):
            v = v.split(
                field.field_info.extra.get(
                    "delimiter", cls.get_list_field_delimiter(field.name)
                )
            )
            v = [item.strip() for item in v if item != ""]
        return v

    return validator(*field_name, allow_reuse=True, pre=True)(split)
```

arthurvd: Great, didn't realize we could automatically turn this inner split function into a class method.
But, .. followup question: now that `cls` is available, the call `cls.get_list_field_delimiter(field.name)` already takes care of **both** the class field's own delimiter (if present) **and** the fallback to the class's own default. So, doesn't that mean that the whole `field.field_info.extra.get` is no longer necessary in the above code?

arthurvd: Done in commits below.

arthurvd (approved):

arthurvd:
@evetion : can you re-review my last 3 commits (in particular the 1st one)? I cannot assign you explicitly as a reviewer here...

PR #227, opened 2022-04-08
#224: Fix fnm serializer to only print filepath item of dict values

### LINKED ISSUE #224: RainfallRunoffModel does not get saved correctly
**Describe the bug**
sobek_3f.fnm does not save correctly

**To Reproduce**
1. run the following code:

`from hydrolib.core.io.fnm.models import RainfallRunoffModel`
`from pathlib import Path`
`fnm = RainfallRunoffModel(filepath=Path("tests/data/input/e02/c11_korte-woerden-1d/dimr_model/rr/Sobek_3b.fnm"))`
`fnm.filepath = "test.fnm"`
`fnm.save()`

3. Notice the file get saved as below, where all objects that are read are saved as a dict-representation of their classes.

**Expected behavior**
Expected all records in test.fnm to be pointing to files relative to the fnm-file.

**Screenshots**
image

 - OS: Windows
 - Version: 0.2.1 branch 223-dimr-rainfallrunoffmodel-fmmodel-are-not-read

**Additional context**

arthurvd:
Fixes #224 
Otherwise, it would print '{'filepath': WindowsPath('3b_nod.tp'), 'node': []}', now it just prints '3b_nod.tp'.

### DIFF
--- a/hydrolib/core/io/fnm/models.py
+++ b/hydrolib/core/io/fnm/models.py
@@ -3,6 +3,7 @@
 from pathlib import Path
 from typing import Callable, Dict, Iterable, Literal, Optional
 
+from pydantic import Field
 from pydantic.types import FilePath
 
 from hydrolib.core.basemodel import FileModel
@@ -20,8 +21,8 @@ class RainfallRunoffModel(FileModel):
 
     # Note that order is defined by the .fnm file type and is used for parsing the data.
     control_file: Path = Path("delft_3b.ini")
-    node_data: Optional[NodeFile]
-    link_data: Optional[LinkFile]
+    node_data: Optional[NodeFile] = Field(None)
+    link_data: Optional[LinkFile] = Field(None)
     open_water_data: Path = Path("3brunoff.tp")
     paved_area_general: Path = Path("paved.3b")
     paved_area_storage: Path = Path("paved.sto")
@@ -31,7 +32,7 @@ class RainfallRunoffModel(FileModel):
     pluvius: Path = Path("pluvius.3b")
     pluvius_general: Path = Path("pluvius.alg")
     kasklasse: Path = Path("kasklass")
-    bui_file: Optional[BuiModel] = None
+    bui_file: Optional[BuiModel] = Field(None)
     verdampings_file: Path = Path("default.evp")
     unpaved_area_general: Path = Path("unpaved.3b")
     unpaved_area_storage: Path = Path("unpaved.sto")
--- a/hydrolib/core/io/fnm/serializer.py
+++ b/hydrolib/core/io/fnm/serializer.py
@@ -2,21 +2,20 @@
 
 import inspect
 from pathlib import Path
-from typing import Dict, Iterable, Optional
+from typing import Dict, Iterable, Optional, Union
 
 
-def _calculate_max_value_length(data: Iterable[Optional[Path]]) -> int:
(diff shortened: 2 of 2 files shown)


hydrolib/core/io/fnm/models.py
```diff
@@ -20,8 +21,8 @@ class RainfallRunoffModel(FileModel):
 
     # Note that order is defined by the .fnm file type and is used for parsing the data.
     control_file: Path = Path("delft_3b.ini")
```
Carsopre, line 23: I would make sure to cover the initialization of this class with tests.

hydrolib/core/io/fnm/serializer.py
```diff
@@ -2,21 +2,20 @@
...
+    return max(map(len, data))
 
 
-def _calculate_offset(value: str, max_len: int) -> str:
-    return (max_len - len(value)) * " "
+def _get_string_value(path_value: Optional[Union[dict, Path]]) -> str:
+    if type(path_value) is dict:
+        value = str(path_value.get("filepath", ""))
+    else:
+        value = str(path_value) if path_value is not None else ""
```
Carsopre, line 16: Personally I think you should be more direct, you are expecting an object type Path, so:
```suggestion
    value: str = "" # Initialize to empty string.
    if type(path_value) is dict:
        value = str(path_value.get("filepath", ""))
    elif isinstance(path, Path):
        value = path_value.resolve()        
```

Of course it would be good to have tests covering this part.

Carsopre: Also as a heads up, I recall the approach of hydrolib was a 'hard fail'. With the above code we do not allow it to crash.

hydrolib/core/io/fnm/serializer.py
```diff
@@ -2,21 +2,20 @@
...
 from pathlib import Path
-from typing import Dict, Iterable, Optional
+from typing import Dict, Iterable, Optional, Union
 
 
-def _calculate_max_value_length(data: Iterable[Optional[Path]]) -> int:
-    sizes = (len(str(v)) for v in data if v is not None)
-    return max(sizes)
+def _calculate_max_value_length(data: Iterable[str]) -> int:
+    return max(map(len, data))
```
Carsopre, line 9: Nice

Carsopre (commented):
I believe the changes are adequate. However i strongly suggest to add a few tests to ensure the future correct behavior of this part of the code.

hydrolib/core/io/fnm/models.py
```diff
@@ -20,8 +21,8 @@ class RainfallRunoffModel(FileModel):
...
-    node_data: Optional[NodeFile]
-    link_data: Optional[LinkFile]
+    node_data: Optional[NodeFile] = Field(None)
+    link_data: Optional[LinkFile] = Field(None)
```
Carsopre, line 25: I would make sure to cover the initialization of this class with tests. (goes as well for bui_file).

hydrolib/core/io/fnm/serializer.py
arthurvd: Thanks @Carsopre , I've implemented this, but also added an ``elif isinstance(path_value, str):`, because that was needed for references to non-existent files in some unit tests. See 7eaebb2.

Covered by test `test_parse_serialize_should_return_same_result()`

PR #231, opened 2022-04-23
chore(devcontainer): fix docker devcontainer behaviour

### LINKED ISSUE #230: devcontainer does not install poetry correctly
When starting a new instance of the devcontainer, the Dockerfile specifies the poetry HYDROLIB-Core environment to be installed in a location separate from the actual repository. This means that the user needs to call `poetry install` separately within the mounted checkout directory. Furthermore, constructing the poetry environment in the separate directory is done for naught, and thus developers have to wait longer than absolutely necessary before they can start developing. 

**To Reproduce**
1. Checkout the HYDROLIB-Core repository
2. Follow the [devcontainer documentation](https://deltares.github.io/HYDROLIB-core/0.2.1/guides/devcontainers/)
3. Start the devcontainer within Visual Studio Code
5. Try and run the test models and note this does not work correctly

**Expected behavior**
The devcontainer should start correctly with an initialised poetry environment which allows the developer to start developing immediately.

**Additional context**
These problems were reported by @rhutten while working with various developers of HKV. An additional request was to update the documentation of the devcontainers to reflect the new state.

BeardedPlatypus:
- Simplify the Dockerfile to only install poetry and other dependencies in the devcontainer
- Extend the devcontainer to run the `poetry install` command after creation of the environment
- Update the devcontainer documentation

Fixes #230

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -21,4 +21,6 @@
 	"mounts": [
 		"source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind"
 	],
+	// This ensures the HYDROLIB environment is properly installed. 
+	"postCreateCommand": "poetry install --no-root --no-interaction --no-ansi"
 }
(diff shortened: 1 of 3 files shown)


arthurvd:
With the latest reply about the Docker performance on https://github.com/Deltares/HYDROLIB-core/issues/230#issuecomment-1160039200, this PR is now ok and I will merge it.

Issue #229, opened 2022-04-21
Add hydrolib.core subpackages for modeldata and geometry


arthurvd:
**What is the need for this task.**
Now that hydrolib.core.io is in place, packages such as D-HyDAMO want to use additional general purpose functionality not related to io. We need new subdirs/subpackages as a starting point.

**What is the task?**
Two new subpackages:
* hydrolib.core.geometry: for grid generation, grid snapping, etc.
* hydrolib.core.modeldata: for convenience functions that translate Pydantic objects to and from dicts/dataframes.

**Additional context**
N/A

PR #257, opened 2022-06-27
#229: Add modeldata and geometry subdirectories

priscavdsluis:


arthurvd (approved):
Changes ok: I removed the __version__ from both __init__.py files, because that was only needed at the `hydrolib.core` level (since that acts as a package on its own with its version number.) Subpackages should not suggest having their own version number.

PR #259, opened 2022-06-27
fix(filemodel): ResolveRelativeMode is incorrectly set when reading a model with 'pathsRelativeToParent' set to false

### LINKED ISSUE #252: Pytest give warnings due to missing files
**Describe the bug**
When executing the command: 'Poetry run pytest' several warnings pop-up. Few of these warnings are related to output, while others result from missing input. Please fix these warnings

**To Reproduce**
In dev containter:
1. Go to terminal
2. Type 'Poetry run pytest'
3. Check the warnings

**Expected behavior**
No warnings should occur when running pytests

**Screenshots**
image

BeardedPlatypus:
Fixes #252

### DIFF
--- a/hydrolib/core/basemodel.py
+++ b/hydrolib/core/basemodel.py
@@ -555,7 +555,7 @@ class FileModel(BaseModel, ABC):
             # Note: the relative mode needs to be obtained from the data directly
             # because self._relative_mode has not been resolved yet (this is done as
             # part of the __init__), however during the __init__ we need to already
-            # have pused the new parent. As such we cannot move this call later.
+            # have pushed the new parent. As such we cannot move this call later.
             relative_mode = self._get_relative_mode_from_data(data)
             context.push_new_parent(filepath.parent, relative_mode)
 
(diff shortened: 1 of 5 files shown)


hydrolib/core/basemodel.py
```diff
@@ -778,6 +784,18 @@ def _relative_mode(self) -> ResolveRelativeMode:
...
 
         Returns:
-            ResolveRelativeMode: The ResolveRelativ
+            ResolveRelativeMode: The ResolveRelativeMode of this FileModel
+        """
+        return ResolveRelativeMode.ToParent
+
+    @classmethod
+    def _get_relative_mode_from_data(cls, data: Dict[str, Any]) -> ResolveRelativeMode:
+        """Gets the ResolveRelativeMode of this FileModel based on the provided data.
```
arthurvd, line 793: I recommend documenting in this most general FileModel implementation of this function that it actually does *not* use the input `data` for getting the ResolveRelativeMode, as the base case that can be overridden by subclasses.

arthurvd (changes requested):
Code ok, please see single remark about documention in the FileModel class.

arthurvd (approved):

Issue #82, opened 2021-10-20
Filenames do not need to exist in .fnm


Carsopre:
I answered this on Discord: "the .fnm file may contain empty values (or a string value "not used"). But: if a file is needed (e.g. "default.bui" when precipitation is enabled), then the filename must be specified (no default filenames are coded in the RR kernel). Finally, some of the specified filenames may not exist on disk, because also output filenames are specified in the input fnm file."

Regarding the .bui file: for now I think we should limit validation of existence of given filenames, because there is no convenient way for HYDROLIB-core to decide whether a filename parameter is for an input file or a not-yet existing output file.

My recommendation: allow all filenames to be empty for now. If they are not empty, do not require them to exist.
Only one exception now is: the buifile, if it is set to a non-empty value. Check for presence (it must exist), parse it, and validate the file by itself.

_Originally posted by @arthurvd in https://github.com/Deltares/HYDROLIB-core/pull/64#discussion_r732504607_

PR #260, opened 2022-06-27
#82: Allow empty file paths in the .fnm file

priscavdsluis:


priscavdsluis:
Note that a couple of tests are currently failing (test_save_and_load_maintains_correct_paths) which should be fixed when #259 is merged back.

BeardedPlatypus:
Issue #82 mentions the following:

> the .fnm file may contain empty values (or a string value "not used").

Do we need to implement this behaviour as part of this issue as well, as I do not see validator dealing with this yet.

BeardedPlatypus (approved):

PR #270, opened 2022-07-07
Remove SonarCloud code smells 

priscavdsluis:


hydrolib/core/io/ini/serializer.py
```diff
@@ -177,7 +177,7 @@ def _serialize_section(self, section: Section) -> Lines:
 
     def _serialize_section_header(self, section_header: str) -> Lines:
         indent = " " * (self.config.section_indent)
-        yield f"{indent}[{section_header}]"
+        return [f"{indent}[{section_header}]"]
```
arthurvd, line 180: It looks correct to me, input is always a single string anyway and never called in a iterable way.
Still, all these `_serialize_*` functions are called in `chain`s, so maybe @BeardedPlatypus can also give his advice.

arthurvd (commented):
All changes ok, but can we please have a quick final check from @BeardedPlatypus on `hydrolib/core/io/ini/serializer.py`?

hydrolib/core/io/ini/serializer.py
BeardedPlatypus: I think this code smell is a false positive, and there is no need for a code change. `Lines` is an alias for `Iterable[str]`. The typing documentation [explicitly states that a generator can be annotated as a `Iterable[YieldType]`](https://docs.python.org/3/library/typing.html#typing.Generator). 

I don't think SonarCloud resolves the `Lines` alias correctly in this instance. I would personally rather mark this as a false positive, then needlessly create a list object every time a section header is read.

arthurvd (approved):

PR #277, opened 2022-07-08
Fix/13 dockertest reenabled

### LINKED ISSUE #13: Fix failing Docker model
includes .nc

arthurvd:
Fixes #13.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,7 +87,7 @@ multi_line_output = 3
 line_length = 88
 
 [tool.pytest.ini_options]
-addopts = "-m \"not plots and not docker\""
+addopts = "-m \"not plots\""
 markers = [
     "plots",
     "docker"
--- a/tests/io/test_docker.py
+++ b/tests/io/test_docker.py
@@ -11,7 +11,7 @@ from tests.utils import test_input_dir, test_output_dir
 from ..utils import test_output_dir
 
 
-@pytest.mark.skip  # skipped as long as its imcomplete
+# @pytest.mark.skip  # skipped as long as its imcomplete
 def test_from_scratch_docker():
     # TODO Make a valid model from scratch
     dimr = DIMR()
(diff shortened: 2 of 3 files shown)


tests/io/test_docker.py
```diff
@@ -11,7 +11,7 @@
 from ..utils import test_output_dir
 
 
-@pytest.mark.skip  # skipped as long as its imcomplete
+# @pytest.mark.skip  # skipped as long as its imcomplete
```
priscavdsluis, line 14: Should this line be removed entirely? And perhaps also the commented out code below?

pyproject.toml
```diff
@@ -87,7 +87,7 @@ multi_line_output = 3
 line_length = 88
 
 [tool.pytest.ini_options]
-addopts = "-m \"not plots and not docker\""
+addopts = "-m \"not plots\""
```
priscavdsluis, line 90: What does this do?

priscavdsluis (changes requested):
Please see remarks

pyproject.toml
priscavdsluis: Thanks! :)

tests/io/test_docker.py
evetion: If it runs, sure! It didn't run before, hence why I skipped it.

evetion:
We could enable these tests again, as they will probably run. But they're incomplete themselves, the actual idea was to generate a completely valid DIMR model and actually *run* it using the FM Docker container automatically. That integration is still missing AFAIK.

tests/io/test_docker.py
arthurvd: I've removed the commented skip in 6c375c4.

evetion:
@arthurvd On second thought, we do run Docker. See the CI step, which mounts the test ouput at `tests/data/output/docker` (meaning we can only run a single test for now!). The actual problem is that we don't get the correct error messages from Docker. It currently fails with:

** ERROR  : Observation cross section file 'ObservationPoints_crs.ini' not found!

PR #281, opened 2022-07-12
feat: support copying generic models

BeardedPlatypus:


docs/topics/loadsave.md
```diff
@@ -23,22 +23,22 @@ own models. As such we will discuss the following topics:
 ## Motivation: HYDROLIB-core should not adjust files when reading and writing
 
 When implementing the model read and write behaviour of HYDROLIB-core, the main 
-motivation is to ensure models are read and written as is. Properties should
+design choise is to ensure models are read and written as is. Properties should
```
arthurvd, line 26: design choice

docs/topics/loadsave.md
```diff
@@ -160,6 +160,24 @@ will be written to disk at the current `save_location`.
 
 Exporting is currently not implemented, but will be as part of [issue #170](https://github.com/Deltares/HYDROLIB-core/issues/170).
 
+## Inheritance hierachy of FileModels: SerializableFileModels and DiskOnlyFileModels
+
+The `FileModel` is the abstract base class for any file models. It provides the logic
```
arthurvd, line 165: any file model

hydrolib/core/basemodel.py
```diff
@@ -607,14 +612,6 @@ def validate(cls: Type["FileModel"], value: Any):
...
 
-    def _load(self, filepath: Path) -> Dict:
-        # TODO Make this lazy in some cases
-        # so it doesn't become slow
-        if filepath.is_file():
-            return self._parse(filepath)
-        else:
-            raise ValueError(f"File: `{filepath}` not found, skipped parsing.")
-
     def save(self, filepath: Optional[Path] = None, recurse: bool = False) -> None:
```
arthurvd, line 615: Should this not also be defined as `@abstractclassmethod`? Or why not?

hydrolib/core/basemodel.py
```diff
@@ -744,6 +723,101 @@ def sync_post(model: BaseModel, acc: FileLoadContext) -> FileLoadContext:
             context.push_new_parent(self._absolute_anchor_path, self._relative_mode)
             traverser.traverse(self, context)
 
+    @property
+    def _relative_mode(self) -> ResolveRelativeMode:
+        """Gets the ResolveRelativeMode of this FileModel.
```
arthurvd, line 728: Docstrings should be in imperative form, right? --> "Get the Resolve[..]"
also in some other places.

hydrolib/core/basemodel.py
```diff
@@ -744,6 +723,101 @@ def sync_post(model: BaseModel, acc: FileLoadContext) -> FileLoadContext:
...
+        Args:
+            filepath (Path): Path to the data to load.
+
+        Returns:
+            Dict: The data stored at filepath
+        """
+        raise NotImplementedError()
+
+
+class SerializableFileModel(FileModel):
```
arthurvd, line 795: Class docstring missing

hydrolib/core/basemodel.py
```diff
@@ -778,31 +852,77 @@ def _get_identifier(self, data: dict) -> Optional[str]:
...
+
+    It implements the FileModel with a void parser and serializer, and a
+    save method which copies the file associated with the FileModel
+    to a new location if it exists.
+
+    We further explicitly assume that when the filepath is None, no
+    file will be written.
+
+    Actual file model implementations *should not* inherit from the
+    DiskOnlyFileModel and instead inherit directly from FileModel.
```
arthurvd, line 868: I think you mean:
"[..] and instead inherit from SerializableFileModel"?

hydrolib/core/basemodel.py
```diff
@@ -778,31 +852,77 @@ def _get_identifier(self, data: dict) -> Optional[str]:
...
-                The unvalidated/parsed data which is fed to the pydantic base model,
-                used to determine the ResolveRelativeMode.
 
-        Returns:
-            ResolveRelativeMode: The ResolveRelativeMode of this FileModel
-        """
-        return ResolveRelativeMode.ToParent
+def validator_set_default_disk_only_file_model_when_none() -> classmethod:
+    """Validator to ensure a default empty DiskOnlyFileModel is provided
+    for an DiskOnlyFileModel which is initialized with None.
```
arthurvd, line 917: for a DiskOnlyFileModel that is initialized with None.

tests/io/rr/test_fnm.py
```diff
@@ -280,3 +320,114 @@ def test_parse_serialize_should_return_same_result():
...
+
+        model.save(filepath=target_path / filepath, recurse=True)
+
+    def assert_correct_subfile(path: Optional[Path]) -> None:
+        assert_file_is_same(path, source_path_parent, target_path)
+
+    assert (target_path / filepath).exists()
+
+    disk_only_file_models = (
+        v for v in dict(model).values() if isinstance(v, DiskOnlyFileModel)
```
arthurvd, line 366: nice :)

hydrolib/core/io/rr/models.py
```diff
@@ -1,220 +1,333 @@
...
 from .serializer import write
 from .topology.models import LinkFile, NodeFile
 
 
-class RainfallRunoffModel(FileModel):
+class ImmutableDiskOnlyFileModel(DiskOnlyFileModel, allow_mutation=False):
+    pass
+
+
+_mappix_paved_area_sewage_storage_name = "pvstordt.his"
```
arthurvd, line 25: Can these not be inside the `RainfallRunoffModel` class? Or why not?

hydrolib/core/io/rr/models.py
```diff
@@ -1,220 +1,333 @@
...
-    lgsi_cachefile: Optional[Path] = None
-    meteo_input_file_rainfall: Optional[Path] = None
-    meteo_input_file_evaporation: Optional[Path] = None
-    meteo_input_file_temperature: Optional[Path] = None
+    _validate_mappix_paved_area_sewage_storage = _validator_mappix_value(
+        "mappix_paved_area_sewage_storage", _mappix_paved_area_sewage_storage_name
+    )
+    mappix_paved_area_sewage_storage: ImmutableDiskOnlyFileModel = Field(
+        ImmutableDiskOnlyFileModel(Path(_mappix_paved_area_sewage_storage_name)),
+        allow_mutation=False,
```
arthurvd, line 166: this is the arguments default value already, but I guess you intentionally want to make that explicit here?

hydrolib/core/io/rr/models.py
```diff
@@ -1,220 +1,333 @@
...
+)
 
 from .meteo.models import BuiModel
 from .parser import read
 from .serializer import write
 from .topology.models import LinkFile, NodeFile
 
 
-class RainfallRunoffModel(FileModel):
+class ImmutableDiskOnlyFileModel(DiskOnlyFileModel, allow_mutation=False):
```
arthurvd, line 21: This deserves some docstring text about its inner workings and purpose.

arthurvd (changes requested):
Hi @BeardedPlatypus, overall I like the new system with two subclasses of abstract FileModel. See my various remarks in this review.

hydrolib/core/basemodel.py
```diff
@@ -778,31 +852,77 @@ def _get_identifier(self, data: dict) -> Optional[str]:
...
+
+    def _save(self) -> None:
+        # The target_file_path contains the new path to write to, while the
+        # _source_file_path contains the original data. If these are not the
+        # same we copy the file and update the underlying source path.
+        target_file_path = self._resolved_filepath
+        if self._can_copy_to(target_file_path):
+            target_file_path.parent.mkdir(parents=True, exist_ok=True)  # type: ignore[arg-type]
+            shutil.copy(self._source_file_path, target_file_path)  # type: ignore[arg-type]
+        self._source_file_path = target_file_path
```
priscavdsluis, line 892: Is it correct to set the _source_file_path even if _can_copy_to returned false?

hydrolib/core/io/rr/models.py
priscavdsluis: I think it is better to omit it, it kinda raises the impression that `allow_mutation=True` is a valid usage of the class, but the class we are using is `ImmutableDiskOnlyFileModel` => it should be implied that `allow_mutation=False`.

hydrolib/core/basemodel.py
BeardedPlatypus: `@abstractclassmethod` indicates that the a subclass will need to define it. If any `@abstractclassmethod` or `@abstractmethod` is not implemented, initializing the subclass will raise an exception. 

In this case, the implementation of save is already defined here, and won't need to be defined in a subclass, which is why it is not abstract. It does rely on several methods which are abstract which need to be implemented. But, because we know that any concrete class has these implemented, `save` will work as expected.

hydrolib/core/basemodel.py
BeardedPlatypus: Well not necessarily, if you want to create a different `FileModel` which might not read the data through a parser / serializer, you might want to inherit from `FileModel` instead of `SerializableFileModel`. It is probably unlikely, so I will update the docstring to reflect this.

hydrolib/core/io/rr/models.py
BeardedPlatypus: > this is the arguments default value already, but I guess you intentionally want to make that explicit here?

Note that this `allow_mutation` is set on the `Field` function, where the default value is `True`. By setting it to `False` we provide a faux immutability to the field, meaning that users cannot assign a new (different) model to the `Field`.
This is different from the `allow_mutation` set on the model, which specifies that all fields of the `ImmutableDiskOnlyFileModel` are faux immutable. By specifying both users cannot adjust the model fields nor the model itself on the `RainfallRunoffModel`. 

As such it is explicitly required to have it here.

hydrolib/core/basemodel.py
BeardedPlatypus: Yes this is intentionally done so. Once we save, we expect the new source path to be equal to the saved location. In my opinion this provides the most consistent behavior. Imagine the following use case:

1. A user creates a new model (without the `DiskOnlyFileModel` backed data).
2. The user saves the model
3. The user creates the files mentioned in the `DiskOnlyFileModel` objects  at the newly saved location
4. The user saves the model in a different location

I would expect the data to then be saved at the new location.

Alternatively, imagine the following:

1. A user creates a new model (without the `DiskOnlyFileModel` backed data).
2. The user saves the model
3. The user creates the files mentioned in the `DiskOnlyFileModel` objects  at the previous saved location
4. The user saves the model in a different location

I would now expect no files to be copied, as there are no files at the new saved location.

This would only work if you do always update the `self._source_file_path`  after saving.

tests/io/rr/test_fnm.py
```diff
@@ -280,3 +320,114 @@ def test_parse_serialize_should_return_same_result():
     reserialized_model = serialize(deserialized_model)
 
     assert reserialized_model == serialized_model
+
+
+def assert_file_is_same(
+    file_path: Optional[Path],
+    read_directory: Path,
+    write_directory: Path,
+) -> None:
```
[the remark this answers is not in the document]
BeardedPlatypus: I probably should have, I'll fix that!

hydrolib/core/io/rr/models.py
arthurvd: ok to leave it as it is

tests/io/rr/test_fnm.py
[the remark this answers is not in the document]
BeardedPlatypus: Actually, this specific function uses `filecmp` and can be used on any file, while the `utils.assert_files_equal` only works for ascii files. I have moved this file to utils as well.

BeardedPlatypus:
The following fields I am not sure about yet, and as such have not been updated:

* `dimr.models`
    * `Component.inputFile` ( I assume this one does not need to be changed)
    * `DIMR.waitFile` (⚠️ is this field name correct to begin with?)
* `ext.models`
    * `Boundary.locationfile`
* `friction.models`
    * `FrictGeneral.frictionvaluesfile`
* `inifield.models`
    * `AbstractSpatialField.datafile`

One additional point, which might need to be addressed in the future, do we want to allow users to disable the copying of files, e.g. with an additional property `should_copy_files` which allows user to overwrite the default behaviour of always copying.

arthurvd:
Hi @BeardedPlatypus, here's my answers:

> The following fields I am not sure about yet, and as such have not been updated:
> 
> * `dimr.models`
>   * `Component.inputFile` ( I assume this one does not need to be changed)

Indeed, all recursive saving of model files is via the underlying FMModel/RRModel, and not in the DIMR class attributes directly.

>   * `DIMR.waitFile` (⚠️ is this field name correct to begin with?)

No need to copy this file: it is a debugging file that is not really part of model input (the field name is correct, though)

> * `ext.models`
>   * `Boundary.locationfile`

Yes, please change this one.

> * `friction.models`
>   * `FrictGeneral.frictionvaluesfile`

Yes, please change this one.

> * `inifield.models`
>   * `AbstractSpatialField.datafile`

Yes, please change this one.

> * `mdu.models`
>   * ` Output.crsfile`: Should this be a list of `DiskOnlyFileModel` objects

Yes, please change this one. And indeed, it's a list.

>   * `Geometry.landboundaryfile`: Should this be a list of `DiskOnlyFileModel` objects

Yes, please change this one. And indeed, it's a list.

> One additional point, which might need to be addressed in the future, do we want to allow users to disable the copying of files, e.g. with an additional property `should_copy_files` which allows user to overwrite the default behaviour of always copying.

Let's first have them test the current functionality.

BeardedPlatypus:
Thanks for the clarification @arthurvd, I'll make sure to wrap those up today.

As a final side note, there are some final paths that I will not replace, as I don't think they fit into the idea of a `DiskOnlyFileModel`:

* `outputdir`
* `waqoutputdir`

I am not sure if we should support copying whole folders. Technically it would probably be possible, but I would recommend moving this to a separate follow up issue.

BeardedPlatypus:
Support for enabling / disabling copying for specific `DiskOnlyFileModel` instances could be trivially added by adding the field:

```python
    should_copy: bool = True
```

To the `DiskOnlyFileModel` and extending the `_save` logic:

```python
    def _save(self) -> None:
        # The target_file_path contains the new path to write to, while the
        # _source_file_path contains the original data. If these are not the
        # same we copy the file and update the underlying source path.
        target_file_path = self._resolved_filepath
        if self.should_copy and self._can_copy_to(target_file_path):
            target_file_path.parent.mkdir(parents=True, exist_ok=True)  # type: ignore[arg-type]
            shutil.copy(self._source_file_path, target_file_path)  # type: ignore[arg-type]
        self._source_file_path = target_file_path
```

tests/utils.py
```diff
@@ -45,3 +46,46 @@ def assert_files_equal(file: Path, reference_file: Path, skip_lines: list = []):
...
+            The input filename relative to input_directory
+        reference_directory (Path):
+            The directory containing the reference file.
+        reference_filepath (Path, optional):
+            The reference filename relative to reference_directory
+            Defaults to ... which is then set to input_filepath.
+    """
+    if input_filepath is None:
+        return
+    if reference_filepath is ...:
```
arthurvd, line 80: Would this not match *anything*? (whereas you mean None/empty string?)

arthurvd (commented):
Hi @BeardedPlatypus , One small question remaining for you.

tests/utils.py
arthurvd: All clear, I overlooked the default argument. Thanks for explaining, review ok now.

arthurvd (approved):

Issue #272, opened 2022-07-07
Refactor get_location_specification_rootvalidator to be less nested and more configurable


priscavdsluis:
The `get_location_specification_rootvalidator` in (hydrolib/core/io/ini/util.py) creates a validator to validate the location specification of certain feature. A valid location specification should be either:

1. A node ID (1D)
2. A branch ID with chainage (1D)
3. An x and y coordinate (2D)

There is no combination possible. Exactly one option should be satisfied.

The validation method has become very nested to, assumably, generate specific warnings for the user. 
Perhaps it is better to validate against an expected valid situation, instead of checking multiple invalid situations.

Moreover, the method should be more configurable. For example, some features cannot have a node ID or branch ID at all. 
The caller should be able to specify which options of the three options they want to validate. We could do this by passing a "configuration" class or a flagged enum as an argument to the `get_location_specification_rootvalidator` function.

Keep in mind that the warnings should still be able to help the user of course. But they should not tell the user what is wrong with the current input, but how a valid input should look like.

Motivations for the refactoring are:

- [SonarCloud complaining about the cognitive complexity](https://sonarcloud.io/project/issues?issues=AYGmvaENuQIJrj8iwgKW&open=AYGmvaENuQIJrj8iwgKW&id=Deltares_HYDROLIB-core)
- [@arthurvd suggesting to use the validator for dam breaks](https://github.com/Deltares/HYDROLIB-core/pull/266/files#r914771466)

PR #289, opened 2022-07-19
Support crs file for observation cross sections

tim-vd-aardweg:


hydrolib/core/io/ini/util.py
```diff
@@ -341,3 +341,117 @@ def validate_coordinates(coord_name: str) -> None:
...
+                )
+
+        number_of_coordinates = get_value(numfield_name)
+        xcoordinates = get_value(xfield_name)
+        ycoordinates = get_value(yfield_name)
+
+        if all_values_are_none():
+            return values
+
+        if coordinates_given_but_none_expected():
```
priscavdsluis, line 438: To reduce the complexicity I would consider replacing the `coordinates_given_but_none_expected()` and `no_coordinates_given_while_expected()` methods with a `some_values_are_none()` method and then using it something like:

```
        if all_values_are_none():
            return values

        if some_values_are_none():
            raise ValueError(
                f"When using coordinates, the fields {numfield_name}, {xfield_name} and {yfield_name} should be given."
            )

        validate_x_and_ycoordinate_number()
```

And the same goes for the coordinate count:
```
        def validate_x_and_ycoordinate_number():
            if len(xcoordinates) != number_of_coordinates or len(ycoordinates) != number_of_coordinates:
                raise ValueError(
                    "Number of x-coordinates and y-coordinates should match number of coordinates."
                )
```

We see that our validation methods often become very nested, because we're trying to be as specific as possible for the user. However, as discussed with @arthurvd , we would like to validate for the correct situation, instead of all the incorrect situations. Because we only have one correct situations, it is easier to validate that instead of validating all the situations that can be wrong.
I hope that makes sense and please let me know what you think.

Please also see #272

hydrolib/core/io/obscrosssection/models.py
```diff
@@ -0,0 +1,86 @@
...
+            "(optional) y-coordinates of the cross section line. "
+            "(number of values = numCoordinates)",
+            alias="yCoordinates",
+        )
+
+    comments: Comments = Comments()
+    _header: Literal["ObservationCrossSection"] = "ObservationCrossSection"
+    name: str = Field(max_length=255, alias="name")
+    branchid: Optional[str] = Field(alias="branchId")
+    chainage: Optional[float] = Field(alias="chaiange")
```
priscavdsluis, line 65:
```suggestion
    chainage: Optional[float] = Field(alias="chainage")
```

hydrolib/core/io/obscrosssection/models.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class ObservationPointCrossSectionModel(INIModel):
+    """
+    The overall observation point crosssection model that contains the contents
+    of one observation point crosssection file.
+    """
+
+    general: ObservationPointCrossSectionGeneral = ObservationPointCrossSectionGeneral()
+    crosssections: List[ObservationPointCrossSection] = []
```
priscavdsluis, line 86: The agreed convention for these lists is to call the field the singular form of the object in the list, so in this case:

```suggestion
    observationpointcrosssection: List[ObservationPointCrossSection] = []
```

(but it should be `observationcrosssection`)

I don't remember why exactly, but I will ask @arthurvd or @BeardedPlatypus when they get back.
There was a specific reason for it.

hydrolib/core/io/obscrosssection/models.py
```diff
@@ -0,0 +1,86 @@
+from typing import List, Literal, Optional
+
+from pydantic.fields import Field
+
+from hydrolib.core.io.ini.models import INIBasedModel, INIGeneral, INIModel
+from hydrolib.core.io.ini.util import (
+    get_location_specification_rootvalidator,
+    get_number_of_coordinates_validator,
+    get_split_string_on_delimiter_validator,
+)
```
priscavdsluis, line 10: A general note: it's not an observation _point_ cross section, but an observation cross section. Please update the names and the comments.

tests/io/ini/test_util.py
```diff
@@ -0,0 +1,81 @@
+from typing import Dict, List, Optional
+
+import pytest
+from pydantic.error_wrappers import ValidationError
+
+from hydrolib.core.io.ini.models import INIBasedModel
+from hydrolib.core.io.ini.util import get_number_of_coordinates_validator
+
```
priscavdsluis, line 8: Nice!!:)

priscavdsluis (changes requested):
Look really nice and well documented!
I do have some remarks. Please take a look and let me know what you think.

hydrolib/core/io/obscrosssection/models.py
tim-vd-aardweg: Good _point_. Fixed.

hydrolib/core/io/ini/util.py
tim-vd-aardweg: I agree that the complexity is a bit high. I was indeed trying to be as specific as possible for the users. But if the users do not care for such specific error messages, your suggestion is indeed way more readable.

If this topic was already discussed before and the conclusion was that it is better to validate for the correct sitatuation, I will do that from now on.

We can always be more specific later if the users decide they want more specific errors!

hydrolib/core/io/obscrosssection/models.py
tim-vd-aardweg: Good catch, fixed.

tests/io/ini/test_util.py
tim-vd-aardweg: Thanks! :D

hydrolib/core/io/obscrosssection/models.py
tim-vd-aardweg: Fixed.

hydrolib/core/io/obscrosssection/models.py
```diff
@@ -0,0 +1,86 @@
...
+
+    comments: Comments = Comments()
+    _header: Literal["ObservationCrossSection"] = "ObservationCrossSection"
+    name: str = Field(max_length=255, alias="name")
+    branchid: Optional[str] = Field(alias="branchId")
+    chainage: Optional[float] = Field(alias="chaiange")
+    numcoordinates: Optional[int] = Field(alias="numCoordinates")
+    xcoordinates: Optional[List[float]] = Field(alias="xCoordinates")
+    ycoordinates: Optional[List[float]] = Field(alias="yCoordinates")
+
```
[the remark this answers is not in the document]
tim-vd-aardweg: So basically we implement this function for all classes that inherit from INIBasedModel? Any other places we have to add this?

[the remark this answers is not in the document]
priscavdsluis: Yes that is correct. 
We also have other implementation, e.g.:
- Each FileModel implementation uses the base implementation which returns the file name.
- Some DIMR components
- XYZPoint returns the coordinates

We basically do it for any object that can be identifyable

priscavdsluis (approved):

Issue #265, opened 2022-06-28
support option to ignore/drop flexible fields in pydantic


xldeltares:
**Is your feature request related to a problem? Please describe.**
In hydrolib-hydromt, we  convert dataframe into pydantic object. The dataframe contains not only the required and optional fields as columns; it also contains flexible fields. Based on the current setting, all these fields are written to the model ini files. Even though these fields wont be read by the kernel, it results in a long ini file with irrelavant information. 

**Describe the solution you'd like**
It would be nice if there is a switch for the user to determine whether or not to keep these flexible fields. 

**Describe alternatives you've considered**
Alternatively, in our script, we include a filter for the data frame for each single object.

**Additional context**
Maarten mentioned there is a possibility in pydantic for such a switch.

BeardedPlatypus:
As mentioned by @xldeltares, I think this is caused by [the `extra` switch set to `Extra.allow`](https://pydantic-docs.helpmanual.io/usage/model_config/#options) in [`INIBasedModel`](https://github.com/Deltares/HYDROLIB-core/blob/main/hydrolib/core/io/ini/models.py#L36). I am not directly sure if this is something we can easily set per model, but it's worth an investigation, from a developer perspective I do see the value of making such behaviour configurable by the end user (with sensible defaults such that in most cases you do not need to do anything).

BeardedPlatypus:
Update on some further discussion with @arthurvd and @priscavdsluis 
We will change the `extra` switch from `allow` to `ignore`. This will ensure the extra data is dropped by default. If a user wants to add an extra field to a model then there are several ways to achieve this. The easiest would probably be to subclass the model and add the extra fields, this should work out of the box with any `INIBasedModel` children, e.g.:

```python
from hydrolib.core.io.mdu.models.py import Restart


class MyCustomRestart(Restart):
    my_property: Literal['hello'] = 'hello'
```

Which can then be written correctly. For reading it might be necessary to update the parents of the extended a class, which could become more cumbersome.

Alternatively we could add additional support for adding fields dynamically if those requirements emerge later on. [Which could for example be achieved through something like this.](https://github.com/samuelcolvin/pydantic/issues/1937) But I think it would be good to only add that, if it turns out to be a common use case to add additional data

priscavdsluis:
@BeardedPlatypus I wonder if the suggested approach is desirable for user. Have you discussed this @arthurvd? Or doesn't it really matter, as long as it work?

PR #291, opened 2022-07-20
#265: Drop extra fields for INIBasedModels

priscavdsluis:


priscavdsluis:
@arthurvd currently there is one test (test_given_structure_text_with_polylinefile_parses_structure``) failing, because the example case contains the following line:
```materialtype               = 1             #1 clay 2 sand, used only in algorithm 1```

Is this also a research keyword that should be documented and added to our model? Because with the implementation in this PR, it will be dropped.

**Comment by arthurvd** [comment]
Hoi @priscavdsluis 
> @arthurvd currently there is one test (`test_given_structure_text_with_polylinefile_parses_structure`) failing, because the example case contains the following line: `materialtype = 1 #1 clay 2 sand, used only in algorithm 1`
> 
> Is this also a research keyword that should be documented and added to our model? Because with the implementation in this PR, it will be dropped.

No, it's an unsupported keyword at the moment, so you can remove it from the testcase.

File: hydrolib/core/io/structure/models.py [hunk 1]
```diff
@@ -60,6 +64,8 @@ class Comments(INIBasedModel.Comments):
     name: str = Field("id", alias="name")
     type: str = Field(alias="type")
 
+    polylinefile: Optional[Path] = Field(None, alias="polylinefile")
```
arthurvd, line 67: Should this be an `Optional[DiskOnlyFileModel]`? Or `Optional[PolyFile]`?

arthurvd (commented):
See my single question.

hydrolib/core/io/structure/models.py
priscavdsluis: I have changed it to a DiskOnlyFile model for now as discussed with you @arthurvd.
Changing it to a PolyFile requires the referenced file to exist. This leads a failing test and we want to look further into how to make it configurable so that we can switch it on or off.

arthurvd (approved):

PR #299, opened 2022-07-29
implement mesh generation for pipes (#294) and fix support different reference system (#298)

### LINKED ISSUE #294: support mesh1d nodes generation for pipes

When a new branch is added to the mehs1d, computational nodes are also generated. The current functionality for generating nodes function will add a minimum of 1 node in the middle of a branch. However, this is not desirable if the new branch is a pipe. For a pipe, computational nodes should only be generated at two branch ends. 
Besides, due to the geometry of pipes being more artificial, a new branch can have two ends that already exist as mesh1d nodes. In this case, the function still tries to add new nodes to the mesh1d, which results in an error saying nodes already exist. 

We would like to have 1) the option not to add computational nodes along the pipes. 2) check if a new node already exists. 

**Describe alternatives you've considered**
Not applicable

**Additional context**
Desired behaviours have been implemented for the delft3dfmpy Deltares-urban branch. But new adaptation is required to add this part to the current hydrolib-core functions.

### LINKED ISSUE #298: incorrect node position was found in mesh1d
**Describe the bug**
In hydrolib.core.io.net.models.Mesh1d, function _node_position is used to find the position of a node if it already exists in the mesh1d.
This function uses np.isclose with default arguments rtol=1e-05, atol=1e-08, (absolute tolerance and relative tolerance) to determine if two nodes are close.  But the rtol is too relaxed, which makes it not applicable for some reference systems. This should be more strict, e.g. using rtol = 0.0.

**To Reproduce**
import numpy as np
n = [1329618.61043459, 5765671.70552735]
a = [1329618.6104345862, 5765631.699667968]
b = [1329618.6104345862, 5765671.705527347]
When do the following:
np.all(np.isclose(a, n)) returns True --> incorrect, because  n[1]-a[1] = 40.00585938245058
np.all(np.isclose(b, n)) returns True --> correct

**Expected behavior**
But using rtol = 0.0:
np.all(np.isclose(a, n, rtol = 0.0)) returns False  --> correct
np.all(np.isclose(b, n, rtol = 0.0)) returns True  --> correct

**Screenshots**
image

 - OS: Windows
 - Version 0.3.0

**Additional context**

xldeltares:
Fixes #294.
Fixes #298.


hydrolib/core/io/net/models.py
```diff
@@ -723,7 +723,7 @@ def _node_position(
         Returns:
             Union[np.int32, None]: The index of the coordinate. None if not found
         """
-        pos = np.where(np.isclose(arrx, x) & np.isclose(arry, y))[0]
+        pos = np.where(np.isclose(arrx, x, rtol=0.0) & np.isclose(arry, y, rtol=0.0))[0]
```
arthurvd, line 726: ~~Do you intentionally set relative tol to zero and keep the absolute tolerance to the default (1e-8)? 
What was going wrong until now?~~ 
Never mind, issue #298 explained it all.

hydrolib/core/io/net/models.py
```diff
@@ -737,6 +737,7 @@ def _add_branch(
         name: str = None,
         branch_order: int = -1,
         long_name: str = None,
+        force_midpoint: bool = True,  # argument to control if a midpoint will be forced on the branch
```
arthurvd, line 740: Please do documentation in a function docstring, not in a comment.

arthurvd (changes requested):
Hi @xldeltares, please check my comments. And a question in general: what did you mean by "different reference system"? Possible coordinate reference system? for the "isclose" fix?

hydrolib/core/io/net/models.py
xldeltares: Indeed I meant "coordinate reference system". :)

hydrolib/core/io/net/models.py
xldeltares: Hi @arthurvd, docstrings are added to the function. please see the commit below.

arthurvd (approved):

PR #300, opened 2022-07-29
Enhancement/238 additional mesh funcs dhydamo

grongen:
Added possibility to take into account structure positions when discretizing 1d mesh. This is needed for hydrolib-dhydamo to function correctly.


priscavdsluis:
I have extracted some function from the `_generate_offsets` function, added some documentation and type hinting.

hydrolib/core/io/net/models.py
```diff
@@ -367,25 +369,167 @@ def __init__(
...
+        if max_dist_to_struc is not None:
+            limits: List[float] = self._update_limits(max_dist_to_struc, limits)
+
+        offsets = self._add_nodes_to_segments(
+            offsets, anchor_pts, limits, mesh1d_edge_length
+        )
+
+        return offsets
+
+    def _update_limits(
```
BeardedPlatypus, line 440: Technically you're not updating if you're returning a new list. Either you would actually update the provided limits, or you could rename it to something like, `_generate_extended_limits`

hydrolib/core/io/net/models.py
```diff
@@ -367,25 +369,167 @@ def __init__(
...
+                )
+
+            dist_to_next_limit = limits[i + 1] - limits[i]
+            if dist_to_next_limit > 2 * max_dist_to_struc:
+                additional.append(
+                    limits[i] + min(2 * max_dist_to_struc, dist_to_next_limit / 2)
+                )
+
+        # Join the limits
+        return sorted(limits + additional)
```
BeardedPlatypus, line 475: I feel like you could probably achieve something more concise with a generator and a stack which yields the next item on the stack and then calculates if an additional node needs to be added something like:

```python
# Note that this untested pseudo-ish code

if len(limits) == 0:
    return []
else:
    return list(_generate_limits)

def _generate_limits(...):    
    stack = list(limits)  # Create a copy to not adjust original limits
    while True:
        curr_item = stack[0]
        yield curr_item
        
        stack = stack[1:]
        if len(stack) > 0:
            if dist_to_next_limit := stack[0] - curr_item > 2 * max_dist_to_struc:
                # Add item to the front of the stack so that it gets yielded next time
                # and we perform the calculation from this point next.
                stack[:0] = curr_item + min(2 * max_dist_to_struc, dist_to_next_limit / 2)
        else:
            # No more items left, time to call it quits
            break
```

Do note that the above code can generate more points than one between two limit nodes. I am not sure if that is by design or a potential bug in the above code. If it is by design, then you might want to adjust the code to yield differently (probably directly yield the extra item instead of pushing it back on the stack).

Further note that because you keep incrementing the nodes, and the original limits are already sorted, your output is also sorted.

hydrolib/core/io/net/models.py
BeardedPlatypus: Also, this could be a static method, which could then be called as `Branch._generate_extended_limits`

hydrolib/core/io/net/models.py
```diff
@@ -367,25 +369,167 @@ def __init__(
...
+            np.ndarray: The array with branch offsets.
+        """
+        # Get upper and lower limits
+        upper_limits = limits[1:]
+        lower_limits = limits[:-1]
+
+        # Determine the segments that are missing a mesh node
+        # Anchor points are added on these segments, such that they will get a mesh node
+        in_range = [
+            ((offsets > lower) & (offsets < upper)).any()
```
BeardedPlatypus, line 502: Shouldn`t this be an `and` instead of a `&`, or is it a numpy thing?

hydrolib/core/io/net/models.py
```diff
@@ -367,25 +369,167 @@ def __init__(
...
+            anchor_pts = sorted(anchor_pts)
+
+            # Generate new offsets
+            offsets = self._generate_1d_spacing(anchor_pts, mesh1d_edge_length)
+
+            # Determine the segments that are missing a grid point
+            in_range = [
+                ((offsets > lower) & (offsets < upper)).any()
+                for lower, upper in zip(lower_limits, upper_limits)
+            ]
```
BeardedPlatypus, line 521: Same as above, might be good to put this in a separate method?

hydrolib/core/io/net/models.py
```diff
@@ -367,25 +369,167 @@ def __init__(
...
+
+        # Check the limits
+        if min(structure_offsets) < 0.0:
+            raise ValueError(
+                f"Distance {min(structure_offsets)} is outside the branch range (0.0 - {self.length})."
+            )
+        if max(structure_offsets) > self.length:
+            raise ValueError(
+                f"Distance {max(structure_offsets)} is outside the branch length (0.0 - {self.length})."
+            )
```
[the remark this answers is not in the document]
priscavdsluis: Will do :)

hydrolib/core/io/net/models.py
```diff
@@ -367,25 +369,167 @@ def __init__(
...
+
+        # Merge limits with start and end of branch
+        limits = [-1e-3] + list(sorted(structure_offsets)) + [self.length + 1e-3]
+
+        # if requested, check if the calculation point are close enough to the structures
+        if max_dist_to_struc is not None:
+            limits: List[float] = self._update_limits(max_dist_to_struc, limits)
```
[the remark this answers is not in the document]
priscavdsluis: Indeed, it is unessecary.

hydrolib/core/io/net/models.py
priscavdsluis: Renamed it

hydrolib/core/io/net/models.py
priscavdsluis: Done

hydrolib/core/io/net/models.py
priscavdsluis: Thanks for the suggestions.  I leave it as is, as the code is not mine and uncovered by tests, so I don't really dare to change it.

tim-vd-aardweg (approved):
Review OK!

Issue #250, opened 2022-06-15
Implement the use of "launch binder"


rhutten:
**What is the need for this task.**
The HYDROLIB team strives to create tutorials, which can be executed through binder. 

**What is the task?**
- create environment.yml or use dockerfile
- make launch buttons at each notebook and at the introduction of the tutorials
- implement binder

BeardedPlatypus:
In #264 a binder Dockerfile has been added. This approach leverages the docker support within Binder. Using either `requirements.txt` or an `environments.yml` proved troublesome, especially in resolving the `hydrolib-core` and `meshkernel` packages. Instead the `Dockerfile` copies the `HYDROLIB-core` repository, and then installs it locally through the `pyproject.toml`. This means we do not need to synchronize any dependencies, which simplifies the process.

In order to create a binding for a jupyter notebook, [the mybinder.org website](https://mybinder.org/) can be used:

![image](https://user-images.githubusercontent.com/5596079/176137877-65a8666e-0004-4c0b-92e1-b293fb961e21.png)

Because the Docker image copies the whole repository, we can specify the jupyter notebook path relative to the repository root. The above settings were used to generate the launch button in #264.

PR #262, opened 2022-06-27
feat: add binder support

BeardedPlatypus:
The following currently works: [Binder](https://mybinder.org/v2/gh/Deltares/HYDROLIB-core/feat/binder-support?labpath=docs%2Ftutorials%2Fbuild_basic_model.ipynb)

However this currently runs on python 3.7 without meshkernel. Trying to add meshkernel, or changing the version to 3.8 or 3.9 leads to crashes in binder. 
@rhutten it might be good to take a look at this together, as I think you have more experience with this.

PR #310, opened 2022-08-18
Feat/binder support

arthurvd:
Was originally intended for #250, but that was later realized via another branch "feat/binder-support-docker".
Question: what should we do with this "feat/binder-support" branch? 
(issue was already reviewed and closed)


arthurvd:
Accidentally opened this one, was already decided in #262  that this non-Docker approach does not work. Closing again (and will delete branch feat/binder-support)

PR #336, opened 2022-09-06
#104: Skip serialization of empty INI properties when configured

### LINKED ISSUE #104: Optional location-related fields should not be written when empty
In the .ext file, locations for Lateral and Boundary objects can be specified in multiple ways, so all these fields are optional.
When an MDU+ext file is saved to disk, also all optional fields are written, which is not an error, but it is unnecessarily confusing, nonetheless.
Example .ext input:
```
[Lateral]
	id = lat_871
	name = lat_871
	type = discharge
	locationType = 1d
	branchId = riv_RS1_909
	chainage = 99.678
	discharge = boundaries.bc

[Boundary]
	quantity = waterlevelbnd
	nodeId = 141133.805000_395446.430000
	forcingFile = boundaries.bc

```
Currently saved output:
```
[Lateral]
id             = lat_871
name           = lat_871
locationType   = 1d
nodeId         =  # <<< May be omitted
branchId       = riv_RS1_909
chainage       = 99.678
numCoordinates =  # <<< May be omitted
xCoordinates   =  # <<< May be omitted
yCoordinates   =  # <<< May be omitted
discharge      = boundaries.bc
type           = discharge

[Boundary]
quantity     = waterlevelbnd
nodeid       = 141133.805000_395446.430000
locationfile =  # <<< May be omitted
forcingfile  = d:\dam_ar\dflowfm_models\HYDROLIB\MGB_FMonly\fm\saved\boundaries.bc
bndwidth1d   =  # <<< May be omitted
bndbldepth   =  # <<< May be omitted
```

**Note** that the same issue arises for Structures (locations) and DamBreaks (waterlevel locations)

priscavdsluis:
Fixes #104

### DIFF
--- a/hydrolib/core/io/mdu/models.py
+++ b/hydrolib/core/io/mdu/models.py
@@ -14,6 +14,7 @@ from hydrolib.core.io.crosssection.models import CrossDefModel, CrossLocModel
 from hydrolib.core.io.ext.models import ExtModel
 from hydrolib.core.io.friction.models import FrictionModel
 from hydrolib.core.io.ini.models import INIBasedModel, INIGeneral, INIModel
+from hydrolib.core.io.ini.serializer import SerializerConfig, write_ini
 from hydrolib.core.io.ini.util import get_split_string_on_delimiter_validator
 from hydrolib.core.io.inifield.models import IniFieldModel
 from hydrolib.core.io.net.models import NetworkModel
@@ -820,3 +821,7 @@ class FMModel(INIModel):
             return ResolveRelativeMode.ToAnchor
         else:
             return ResolveRelativeMode.ToParent
+
+    def _serialize(self, _: dict) -> None:
+        config = SerializerConfig(skip_empty_properties=False)
+        write_ini(self._resolved_filepath, self._to_document(), config=config)
--- a/tests/io/test_ini.py
+++ b/tests/io/test_ini.py
@@ -1607,7 +1607,7 @@ class TestSerializer:
                     property_indent=4,
                     datablock_indent=6,
                     datablock_spacing=4,
-                    skip_empty_properties=False
+                    skip_empty_properties=False,
                 ),
                 [
                     "[header]",
@@ -1631,12 +1631,9 @@ class TestSerializer:
                     property_indent=4,
                     datablock_indent=6,
                     datablock_spacing=4,
-                    skip_empty_properties=True
+                    skip_empty_properties=True,
                 ),
-                [
-                    "[header]",
-                    "    key1 = value # comment1"
-                ],
+                ["[header]", "    key1 = value # comment1"],
             ),
         ],
     )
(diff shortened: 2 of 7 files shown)


hydrolib/core/io/mdu/models.py
```diff
@@ -820,3 +821,7 @@ def _get_relative_mode_from_data(cls, data: Dict[str, Any]) -> ResolveRelativeMo
             return ResolveRelativeMode.ToAnchor
         else:
             return ResolveRelativeMode.ToParent
+
+    def _serialize(self, _: dict) -> None:
+        config = SerializerConfig(skip_empty_properties=False)
+        write_ini(self._resolved_filepath, self._to_document(), config=config)
```
tim-vd-aardweg, line 827: Visual Studio gives the following error for `self._resolved_filepath`:

`Argument of type "Path | None" cannot be assigned to parameter "path" of type "Path" in function "write_ini"
  Type "Path | None" cannot be assigned to type "Path"
    Type "None" cannot be assigned to type "Path"`

I think it is because `_resolved_filepath` is `Optional` and `write_ini()` expects a non-optional path.

tim-vd-aardweg (changes requested):

tests/io/test_ini.py
```diff
@@ -1583,6 +1596,49 @@ def test_serialize_datablock(
...
+                    "    key3 =       # comment3",
+                ],
+            ),
+            (
+                Section(
+                    header="header",
+                    content=[
+                        Property(key="key1", value="value", comment="comment1"),
+                        Property(key="key2", value="", comment="comment2"),
+                        Property(key="key3", value=None, comment="comment3"),
```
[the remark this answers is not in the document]
priscavdsluis: Done!

hydrolib/core/io/mdu/models.py
arthurvd: Is that so? I see such a check only in `ParsableFileModel._serialize()`, but that will not be called if a subclass overrides that method, for example in `INIModel._serialize()`.

I think the place that now makes sure everything runs well is this:
```
class FileModel(BaseModel, ABC):
# ...
    def _save_instance(self) -> None:
        if self.filepath is None:
```

Should we not add documentation that `_serialize()` should always only be called via `save()` and never directly?

arthurvd (commented):
@priscavdsluis: almost done, please see one remaining comment.

hydrolib/core/io/mdu/models.py
arthurvd: @priscavdsluis wrote that the latter is implicitly so, because `_serialize()` is a private function. That is true for end-user APIdocs. But amongst core-developers, it is good to know that *other* private functions should probably also not call `_serialize` directly. Maybe some one-liner docs in the function body?

priscavdsluis: I added some docs and will close this PR.

PR #349, opened 2022-10-03
Change AutoStart type from bool to int

veenstrajelmer:
AutoStart type should be int instead of bool, since 2 is also a allowed value according to the manual: https://content.oss.deltares.nl/delft3d/manuals/D-Flow_FM_User_Manual.pdf#section.A.1

### DIFF
--- a/hydrolib/core/io/mdu/models.py
+++ b/hydrolib/core/io/mdu/models.py
@@ -30,7 +30,7 @@ class General(INIGeneral):
     version: str = Field("1.2.94.66079M", alias="version")
     filetype: Literal["modelDef"] = Field("modelDef", alias="fileType")
     fileversion: str = Field("1.09", alias="fileVersion")
-    autostart: bool = Field(False, alias="autoStart")
+    autostart: int = Field(0, alias="autoStart")
     pathsrelativetoparent: bool = Field(False, alias="pathsRelativeToParent")
 
 
--- a/hydrolib/core/io/mdu/models.py
+++ b/hydrolib/core/io/mdu/models.py
@@ -24,13 +24,23 @@ from hydrolib.core.io.structure.models import StructureModel
 from hydrolib.core.io.xyz.models import XYZModel
 
 
+class AutoStartOption(IntEnum):
+    """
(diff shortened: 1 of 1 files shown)


hydrolib/core/io/mdu/models.py
```diff
@@ -30,7 +30,7 @@ class General(INIGeneral):
     version: str = Field("1.2.94.66079M", alias="version")
     filetype: Literal["modelDef"] = Field("modelDef", alias="fileType")
     fileversion: str = Field("1.09", alias="fileVersion")
-    autostart: bool = Field(False, alias="autoStart")
+    autostart: int = Field(0, alias="autoStart")
```
priscavdsluis, line 33: We also like to use enums for limited options such as in this case. 
This comes with the advantage that we can also document the several options. For example that `0` is `no`, `1` = `autostart` and `2` = `autostartstop`. 

I _think_ (not sure) that we get also "free" validation on the values, for example that 3 or -1 is not valid.


```
class AutoStartOption(IntEnum):
    """
    Enum class containing the valid values for the AutoStart
    attribute in the [General][hydrolib.core.io.mdu.models.General] class.
    """
    no=0
    autostart=1
    autostartstop=2
```


```
autostart: AutoStartOption = Field(AutoStartOption.no, alias="autoStart")
```

priscavdsluis (changes requested):
@veenstrajelmer, I reviewed the changes. Please let me know what you think.

veenstrajelmer:
Great, looks good.

hydrolib/core/io/mdu/models.py
```diff
@@ -662,7 +672,7 @@ class Processes(INIBasedModel):
     thetavertical: Optional[float] = Field(0.0, alias="ThetaVertical")
     dtprocesses: Optional[float] = Field(0.0, alias="DtProcesses")
     dtmassbalance: Optional[float] = Field(0.0, alias="DtMassBalance")
-    processfluxintegration: Optional[float] = Field(
+    processfluxintegration: Optional[ProcessFluxIntegration] = Field(
```
priscavdsluis, line 675: Good catch!

PR #356, opened 2022-10-04
T3D header not correct

### LINKED ISSUE #317: t3D header not correct yet
**Describe the bug**
Some issues with t3D header:
Missing:
 - vertical position 1/2/3/n is not supported, should be added to QuantityUnitPair (or new object)
 - timeInterpolation not written to file for T3D blocks
 - verticalPositionType "ZDatum" but is not available in VerticalPositionType class
 - verticalPositionType "ZSurf" is missing (_newly added to this issue, Sep 29 '22_)
Inconsistencies:
 - "verticalPositions" is written, while "Vertical position specification" is expected (in 2D3D manual). Does both work?
   **Answer:** in kernel: yes, in GUI: no. And we should move to 1 consistent keyword. Decision (Oct 13, 2022): support only the latest kernel keywords  in the model fields + support spaced old keywords for compatibility via a pre root_validators, e.g., "Vertical position specification"
  (Pydantic already turns spaced fields into underscore fields, e.g., "vertical_position_specification")
  Background for D-HYDRO developers, see internal issue UNST-5801.
 - "verticalPositionType " is written, while "Vertical position type" is expected (in 2D3D manual). Does both work?
  **Answer:** same as previous one. Support "vertPositionType" and backwards compability for "Vertical position type".

In summary, these are the correct keywords:
* `timeInterpolation`
* `vertPositionType`
* `vertPositions`
* `vertInterpolation`
* `vertPositionIndex`

**To Reproduce**
```
from pathlib import Path
import numpy as np

from hydrolib.core.io.bc.models import (
    ForcingModel,
    QuantityUnitPair,
    TimeInterpolation,
    T3D,
    VerticalPositionType,
)

bc_south = ForcingModel()
depth_list = [-2000,-1000,-500,-100]
times_list = np.array([0,60,120,180,240,300,360,420])
datablock = np.random.uniform(low=-40, high=130.3, size=(len(times_list),len(depth_list)))
datablock_incltime = np.concatenate([times_list[:,np.newaxis],datablock],axis=1)
list_QUP_perlayer = [QuantityUnitPair(quantity='salinity', unit='1e-3') for iL in range(len(depth_list))] #TODO REPORT: verticalposition 1/2/3/n is not supported
ts_one = T3D(name='3d_bnd_0001',
             verticalpositions=depth_list, #TODO: should be "Vertical position specification = [..]" but is verticalPositions = [..]" (both possible?)
             verticalInterpolation='linear',
             verticalPositionType=VerticalPositionType('ZBed'), #TODO: should be "Vertical position type = zdatum" but is "verticalPositionType = ZBed" (zdatum is niet beschikbaar)
             quantityunitpair=[QuantityUnitPair(quantity="time", unit="seconds since 2022-01-01 00:00:00 +00:00")]+list_QUP_perlayer,
             timeinterpolation=TimeInterpolation.linear, #TODO: not passed on to bc file
             datablock=datablock_incltime.tolist(), 
             )
bc_south.forcing.append(ts_one)
bc_south.save(filepath=Path("steric_south2.bc"))
```

**Expected behavior**
described in description

 - OS: Windows, Anaconda, Spyder
 - Version: 0.3.0

tim-vd-aardweg:
Fixes #317


hydrolib/core/io/bc/models.py
```diff
@@ -236,6 +291,152 @@ class T3D(ForcingBase):
...
+    )
+
+    @root_validator(pre=True)
+    def _validate_quantityunitpair(cls, values):
+        super()._validate_quantityunitpair(values)
+
+        quantityunitpairkey = "quantityunitpair"
+        quantityunitpairs = values[quantityunitpairkey]
+
+        T3D._validate_that_first_unit_is_time_and_has_no_verticalpositionindex(
```
arthurvd, line 305: Is it necessary (or style?) to include our own class name here?

hydrolib/core/io/bc/models.py
```diff
@@ -236,6 +291,152 @@ class T3D(ForcingBase):
...
     )
+    _timeinterpolation_validator = get_enum_validator(
+        "timeinterpolation", enum=TimeInterpolation
+    )
+
+    @root_validator(pre=True)
+    def _validate_quantityunitpair(cls, values):
+        super()._validate_quantityunitpair(values)
+
+        quantityunitpairkey = "quantityunitpair"
```
arthurvd, line 302: this var is only used once, on the next line. Maybe just have the "quantityunitpair" directly in that dict-lookup and remove this variable?

hydrolib/core/io/bc/models.py
```diff
@@ -236,6 +291,152 @@ class T3D(ForcingBase):
...
+        verticalpositionindexes = values.get("verticalpositionindex")
+        number_of_verticalpositions = len(values["verticalpositions"])
+
+        if verticalpositionindexes is None:
+            T3D._validate_that_all_non_time_quantityunitpairs_have_valid_verticalpositionindex(
+                quantityunitpairs, number_of_verticalpositions
+            )
+            return values
+
+        T3D._validate_verticalpositionindexes_and_update_quantityunitpairs(
```
arthurvd, line 318: For clarity maybe put this in an `else` clause? (now it really relies on the previous `return` to not come here.)

hydrolib/core/io/bc/models.py
```diff
@@ -236,6 +291,152 @@ class T3D(ForcingBase):
...
+            ):
+                raise ValueError(
+                    f"Vertical position index should be between 1 and {number_of_vertical_positions}"
+                )
+
+    @staticmethod
+    def _is_valid_verticalpositionindex(
+        verticalpositionindex: int, number_of_vertical_positions: int
+    ) -> bool:
+        one_based_index_offset = 1
```
arthurvd, line 419: not super useful? If (hypothetically) later it would be changed to `= 0` then still the expression below would need modification ` < number_of_vertical_positions`, instead of `<=`.
So, maybe directly put the `1` in the expression below?

hydrolib/core/io/bc/models.py
```diff
@@ -236,6 +291,152 @@ class T3D(ForcingBase):
...
+    ) -> None:
+        if verticalpositionindexes is None:
+            raise ValueError("verticalpositionindex is not provided")
+
+        number_of_quantityunitpairs = len(quantityunitpairs)
+        number_of_verticalpositionindexes = len(verticalpositionindexes)
+
+        if number_of_verticalpositionindexes == number_of_quantityunitpairs:
+            T3D._validate_verticalpositionindexes_and_add_them_to_quantityunitpairs(
+                verticalpositionindexes[1:],
```
arthurvd, line 366: Which case are you handling here? When are these numbers equal, and why are you then skipping element 0?

hydrolib/core/io/bc/models.py
```diff
@@ -266,16 +472,16 @@ class ForcingModel(INIModel):
...
     """
 
     general: ForcingGeneral = ForcingGeneral()
+    """ForcingGeneral: `[General]` block with file metadata."""
+
     forcing: List[ForcingBase] = []
+    """List[ForcingBase]: List of `[Forcing]` blocks for all forcing
+    definitions in a single .bc file. Actual data is stored in
+    forcing[..].datablock from [hydrolib.core.io.ini.models.DataBlockINIBasedModel.datablock] 
+    or [hydrolib.core.io.ini.models.DataBlockINIBasedModel]."""
```
arthurvd, line 484: Why this "or.. " sentence?

arthurvd (changes requested):
@tim-vd-aardweg : please see my recommendations and some questions.

veenstrajelmer:
I have tested the code and after changing verticalPositionType to verticalpositiontype (there was no warning/error on supplying a wrong key, only the actual key was missing), I got it working. It seems to work, but when I put it in a Boundary, there is an error:
```
ValidationError: 3 validation errors for Boundary
forcingFile -> salinity_DCSM-FM_OB_all_20181108_nocomments.bc -> forcing -> 0
  Function of ForcingBase with name=DCSM-FM_OB_all_20181108_0001 and function=t3d is not recognized. (type=value_error)
forcingFile -> salinity_DCSM-FM_OB_all_20181108_nocomments.bc -> forcing -> 1
  Function of ForcingBase with name=DCSM-FM_OB_all_20181108_0002 and function=t3d is not recognized. (type=value_error)
forcingFile -> salinity_DCSM-FM_OB_all_20181108_nocomments.bc -> forcing -> 2
  Function of ForcingBase with name=DCSM-FM_OB_all_20181108_0003 and function=t3d is not recognized. (type=value_error)
```
I searched but could not find the code where this error is raised,

veenstrajelmer:
I have tested the code and after changing verticalPositionType to verticalpositiontype (there was no warning/error on supplying a wrong key, only the actual key was missing), I got it working. It seems to work, but when I put it in a Boundary, there is an error:
```
ValidationError: 3 validation errors for Boundary
forcingFile -> salinity_DCSM-FM_OB_all_20181108_nocomments.bc -> forcing -> 0
  Function of ForcingBase with name=DCSM-FM_OB_all_20181108_0001 and function=t3d is not recognized. (type=value_error)
forcingFile -> salinity_DCSM-FM_OB_all_20181108_nocomments.bc -> forcing -> 1
  Function of ForcingBase with name=DCSM-FM_OB_all_20181108_0002 and function=t3d is not recognized. (type=value_error)
forcingFile -> salinity_DCSM-FM_OB_all_20181108_nocomments.bc -> forcing -> 2
  Function of ForcingBase with name=DCSM-FM_OB_all_20181108_0003 and function=t3d is not recognized. (type=value_error)
```
I searched but could not find the code where this error is raised,

veenstrajelmer (approved):
Tested it with the testcase from https://github.com/Deltares/HYDROLIB-core/issues/317. All reported issues are solved

hydrolib/core/io/bc/models.py
```diff
@@ -32,9 +37,16 @@
...
+    """str: Linear interpolation between vertical positions."""
+
     log = "log"
+    """str: Logarithmic interpolation between vertical positions (e.g. vertical velocity profiles)."""
+
     block = "block"
+    """str: Equal to the value at the directly lower specified vertical position."""
 
 
 class VerticalPositionType(str, Enum):
```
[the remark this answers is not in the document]
tim-vd-aardweg: Done!

hydrolib/core/io/bc/models.py
```diff
@@ -220,11 +264,22 @@ class T3D(ForcingBase):
...
+    """List[float]: The specification of the vertical positions."""
 
-    verticalpositions: List[float] = Field(alias="verticalPositions")
-    verticalinterpolation: VerticalInterpolation = Field(alias="verticalInterpolation")
-    verticalpositiontype: VerticalPositionType = Field(alias="verticalPositionType")
+    verticalinterpolation: VerticalInterpolation = Field(alias="Vertical Interpolation")
+    """VerticalInterpolation: The type of vertical interpolation."""
+
+    verticalpositiontype: VerticalPositionType = Field(alias="Vertical Position Type")
+    """VerticalPositionType: The vertical position type."""
```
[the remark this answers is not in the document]
tim-vd-aardweg: Done.

hydrolib/core/io/bc/models.py
tim-vd-aardweg: It is necessary. Since the method that follows is a (private) static method. The alternative would be to add these methods that are currently (private) static as local functions in the root validator function `_validate_quantityunitpair` itself.

I don't have a _strong_ preference for either option, but if I had to pick I'd got with the static methods since it makes the root validator `_validate_quantityunitpair` less clogged and more readable.

hydrolib/core/io/bc/models.py
tim-vd-aardweg: Good point! I stole this from the base implementation, where it is in fact used more often. Since it is only used once here, I will remove this line.

tim-vd-aardweg: Done

hydrolib/core/io/bc/models.py
tim-vd-aardweg: We discussed this via Teams.

The idea here was that the user could specify the `verticalpositionindex` in two ways. The first way would be an array with a length equal to the number of quantities/units given. The second option would be an array with a length equal to the quantities/units given - 1.

In the first case (same length arrays) the user has specified a `verticalpositionindex` for all quantities, including `time`. Since `time` cannot have a `verticalpositionindex`, the code you highlighted skips the first entry of the `verticalpositionindexes`, bascially ignoring the value the user entered.

We discussed that this is not what we want and that I will change it to the following:
The user should always provide an array of `verticalpositionindex` that is 1 less than the number of quantities/units provided by the user. If the user specifies more (or less) values, a `ValueError` will be raised.

So the only valid option for the user would be:
`
quantities = ["time", "randomQuantity1", "randomQuantity2"]
units = ["randomUnit", "randomUnit", "randomUnit"]
verticalpositionindex = [1, 2]
`

hydrolib/core/io/bc/models.py
tim-vd-aardweg: Fixed.

hydrolib/core/io/bc/models.py
tim-vd-aardweg: It is to clarify that we are dealing with a one-based index instead of the usual zero-based index.

hydrolib/core/io/bc/models.py
tim-vd-aardweg: The `else` clause will add unnecesarry indentation to our code. So I prefer not to add the `else` clause.

hydrolib/core/io/bc/models.py
tim-vd-aardweg: Fixed!

tim-vd-aardweg:
Implemented review comments and fixed a small bug.

hydrolib/core/io/bc/models.py
```diff
@@ -220,11 +264,22 @@ class T3D(ForcingBase):
...
     factor: float = Field(1.0, alias="factor")
+    """float: All values in the table are multiplied with the factor. Defaults to 1.0."""
+
+    verticalpositions: List[float] = Field(alias="Vertical Position Specification")
+    """List[float]: The specification of the vertical positions."""
 
-    verticalpositions: List[float] = Field(alias="verticalPositions")
-    verticalinterpolation: VerticalInterpolation = Field(alias="verticalInterpolation")
-    verticalpositiontype: VerticalPositionType = Field(alias="verticalPositionType")
+    verticalinterpolation: VerticalInterpolation = Field(alias="Vertical Interpolation")
```
veenstrajelmer, line 277: default 'linear' toevoegen

tim-vd-aardweg: Fixed

tim-vd-aardweg:
Discussed with @arthurvd that we will use the following keywords for reading and writing:
- verticalPositionSpecification (we will not support Vertical Position Specification or verticalPositions)
- verticalPositionType (we will not support Vertical Position Type)
- verticalPosition (we will not support Vertical Position or verticalPositionIndex)

arthurvd:
> Issue is currently on hold while we internally discuss the keywords once more.

Apologies for al the unclarity: after discussion with @veenstrajelmer , we propose to directly use the latest official keywords:
https://issuetracker.deltares.nl/browse/UNST-5801, under "New keyword kernel (2019)", that is:
* `timeInterpolation`
* `vertPositionType`
* `vertPositions`
* `vertInterpolation`
* `vertPositionIndex`

hydrolib/core/io/bc/models.py
```diff
@@ -220,23 +271,177 @@ class T3D(ForcingBase):
...
+
+        quantityunitpairs = values["quantityunitpair"]
+
+        T3D._validate_that_first_unit_is_time_and_has_no_verticalposition(
+            quantityunitpairs
+        )
+
+        verticalpositions = values.get("vertpositions")
+        if verticalpositions is None:
+            raise ValueError("vertpositions is not provided")
```
priscavdsluis, line 331: Convention is to use the lower camel case names of the keywords: `vertPositions` & `vertPositionIndex`

hydrolib/core/io/bc/models.py
```diff
@@ -220,23 +271,177 @@ class T3D(ForcingBase):
...
+        if quantityunitpairs[0].quantity.lower() != "time":
+            raise ValueError("First quantity should be `time`")
+        if quantityunitpairs[0].vertpositionindex is not None:
+            raise ValueError("`time` quantity cannot have vertical position index")
+
+    @staticmethod
+    def _validate_that_all_non_time_quantityunitpairs_have_valid_verticalpositionindex(
+        quantityunitpairs: List[QuantityUnitPair], maximum_verticalpositionindex: int
+    ) -> None:
+        for quantityunitpair in quantityunitpairs:
```
priscavdsluis, line 367:
```suggestion
        for quantityunitpair in quantityunitpairs[1:]:
```

and then you can remove

```
            if quantity == "time":
                continue
```

hydrolib/core/io/bc/models.py
```diff
@@ -220,23 +271,177 @@ class T3D(ForcingBase):
...
+                )
+
+    @staticmethod
+    def _validate_verticalpositionindexes_and_update_quantityunitpairs(
+        verticalpositionindexes: List[int],
+        number_of_verticalpositions: int,
+        quantityunitpairs: List[QuantityUnitPair],
+    ) -> None:
+        if verticalpositionindexes is None:
+            raise ValueError("vertpositionindex is not provided")
```
priscavdsluis, line 387:
```suggestion
            raise ValueError("vertPositionIndex is not provided")
```

arthurvd: @priscavdsluis : I'm fine with this (I even like it a bit better). Only we are not consistent throughout the code: sometimes we print the class/Pydantic fields names (lowercased), and sometimes the alias. Question is mainly: are we adressing the end-user or a core developer? (I vote: end user)

So, I'll commit this suggested change, but we should do it in more places.

arthurvd (commented):
Will continue a separate review directly after this for Tim's later commits of this day. Those will contain important cleanups for the field name validators

hydrolib/core/io/ini/util.py
```diff
@@ -462,3 +462,28 @@ def is_valid_coordinates_with_num_coordinates_specification() -> bool:
         raise ValueError(error)
 
     return root_validator(allow_reuse=True)(validate_location_specification)
+
+
+def get_key_renaming_root_validator(keys_to_rename: Dict[str, List[str]]):
```
arthurvd, line 467: Yes! Like this a lot :-)

arthurvd (approved):
I'm ok after Prisca's earlier comments have been resolved.

hydrolib/core/io/ini/util.py
tim-vd-aardweg: Thanks! :)

hydrolib/core/io/bc/models.py
```diff
@@ -269,58 +357,62 @@ class T3D(ForcingBase):
...
     )
     """VerticalInterpolation: The type of vertical interpolation. Defaults to linear."""
 
-    verticalpositiontype: VerticalPositionType = Field(alias="verticalPositionType")
+    vertpositiontype: VerticalPositionType = Field(alias="vertPositionType")
     """VerticalPositionType: The vertical position type of the verticalpositions values."""
 
     timeinterpolation: TimeInterpolation = Field(
-        TimeInterpolation.linear, alias="timeInterpolation"
+        TimeInterpolation.linear.value, alias="timeInterpolation"
```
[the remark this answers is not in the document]
tim-vd-aardweg: Here, it cannot. If we leave it out here, it will write `TimeInterpolation.linear` to the files.

hydrolib/core/io/ini/util.py
priscavdsluis, line 467: I think this functionality can also be used for non-ini files. 
We could move this to a different namespace in another issue, since it is not important now. 

Also, I like the current generic approach.

I think in the future we could introduce a dedicated class of which an instance can be passed to this function.
We should also try to make use the semantic versioning from the files to do backwards compatibility, this will make it more reliable over the lifespan of hydrolib-core.

hydrolib/core/io/ini/util.py
```diff
@@ -462,3 +462,28 @@ def is_valid_coordinates_with_num_coordinates_specification() -> bool:
...
+    """
+
+    def rename_keys(cls, values: Dict) -> Dict:
+        for current_keyword, old_keywords in keys_to_rename.items():
+            if values.get(current_keyword) is not None:
+                continue
+
+            for old_keyword in old_keywords:
+                if values.get(old_keyword) is not None:
+                    values[current_keyword] = values.get(old_keyword)
```
priscavdsluis, line 483:
```suggestion
                if (value := values.get(old_keyword)) is not None:
                    values[current_keyword] = value
```

You can use the walrus operator for this (maybe a more specific name, but I couldn't come up with anything :))

hydrolib/core/io/ini/util.py
```diff
@@ -462,3 +462,28 @@ def is_valid_coordinates_with_num_coordinates_specification() -> bool:
...
+    Gets a root validator that renames the provided keys to support backwards compatibility.
+
+    Args:
+            keys_to_rename Dict[str, List[str]]: Dictionary of keys and a list of old keys that
+            should be converted to the current key.
+    """
+
+    def rename_keys(cls, values: Dict) -> Dict:
+        for current_keyword, old_keywords in keys_to_rename.items():
+            if values.get(current_keyword) is not None:
```
priscavdsluis, line 478:
```suggestion
            if current_keyword in values:
```

Just another way to do it, but should not be changed.

hydrolib/core/io/bc/models.py
```diff
@@ -220,23 +271,177 @@ class T3D(ForcingBase):
...
+
+    timeinterpolation: TimeInterpolation = Field(
+        TimeInterpolation.linear.value, alias="timeInterpolation"
+    )
+    """TimeInterpolation: The type of time interpolation. Defaults to linear."""
+
+    _key_renaming_root_validator = get_key_renaming_root_validator(
+        {
+            "timeinterpolation": ["time_interpolation"],
+            "vertpositions": ["vertical_position_specification"],
```
priscavdsluis, line 298: I don't really get these keywords?
To be backwards compatible, shouldn't for examle  `verticalpositions` be part of the old keys, since that was our old key?
Also are spaces replaced with underscores? Is that the reason why it is `vertical_position_specification` instead of `vertical position specification`?

tim-vd-aardweg: The idea here is that the first keyword (i.e the key of the dictionary) is the current keyword that should be used. What follows (the value belonging to a key) is a list of older keywords. So we can look for multiple older keywords and rename them to the current version of the keyword. Do you have any idea how we can make this more clear?

I will ask @arthurvd if we should add more old keys here, like the one you mentioned.

Regarding your question about spaces and underscores: Indeed, the keyword specified in the file can be `Vertical Position Specification`, for example. At first I thought pydantic replaced the spaces with underscores, but after a little more digging in our code, it turns out we convert those spaces to underscores in our `to_key` function that constructs a key name from a given field name. So that's why I had to specify the keyword with underscores.

Do you think it would be more clear if we used `to_key("Vertical Position Specification")` here instead?

hydrolib/core/io/ini/util.py
tim-vd-aardweg: I like your way better. It's a little cleaner. I will change it!

hydrolib/core/io/ini/util.py
tim-vd-aardweg: Cool, I didn't know. I have update the code.

hydrolib/core/io/ini/util.py
tim-vd-aardweg: Agreed. I talked about this validator being useful for non-ini files with Arthur. We decided we can make it more generic when we need it. So for now we wanted to just have it for the ini files.

Also completely agree with your second point. It would cleaner if we pass an instance of a (new, to be created) class to this function.

One of the problems of the versioning of the files is that they are not consistant. Take a look, for example, at the documentation in the 1D2D and 2D3D manuals and look for the .bc file documentation. Both have fileversion 1.01, but they use different keywords :(

Shall I make a new issue for improving the this validator? Or do we just create a new issue later, when/if we decide to work on it?

hydrolib/core/io/bc/models.py
tim-vd-aardweg: Yes, I like it. Done!

hydrolib/core/io/bc/models.py
```diff
@@ -220,23 +271,177 @@ class T3D(ForcingBase):
...
+        quantityunitpairs = values["quantityunitpair"]
+
+        T3D._validate_that_first_unit_is_time_and_has_no_verticalposition(
+            quantityunitpairs
+        )
+
+        verticalpositions = values.get("vertpositions")
+        if verticalpositions is None:
+            raise ValueError("vertpositions is not provided")
+        verticalpositionindexes = values.get("vertpositionindex")
```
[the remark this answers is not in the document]
tim-vd-aardweg: Done

hydrolib/core/io/bc/models.py
tim-vd-aardweg: Done

hydrolib/core/io/bc/models.py
tim-vd-aardweg: Discussed this with Prisca. We will not use the `to_key` function here.

hydrolib/core/io/ini/util.py
tim-vd-aardweg: We will create a follow-up issue in which we will think about our requirements for backwards compatibility. After that, we can improve this implementation. For now: no action needed.

priscavdsluis (approved):

PR #359, opened 2022-10-07
Update parser.py

veenstrajelmer:
This warning is mainly inconvenient. The default formatting of polyline files contains whitespaces, so it is printed for every polyobject in every polyfile that is read. I could not find a more convenient way of disabling this warning then removing this part of the code.


priscavdsluis (approved):

priscavdsluis:
I will close this PR. I have created PR #409, because some more things had to be adjusted.

PR #367, opened 2022-10-11
341: Correct full file path casing upon initializing model

priscavdsluis:
Added full file path case resolving for Windows, Linux and MacOS.
The resolving works differently for each OS.
For MacOS this can be performance heavy, depending on your file structure.


tests/test_basemodel.py
```diff
@@ -24,6 +24,11 @@
 _external_path = test_output_dir / "test_save_and_load_maintains_correct_paths_external"
 
 
+def runs_from_docker() -> bool:
+    """Check to see if we are running from within docker."""
+    return Path("/.dockerenv").exists()
```
tim-vd-aardweg, line 29: Nice, for now this should be fine! :)

tim-vd-aardweg (approved):

tim-vd-aardweg:
Review OK!

Issue #329, opened 2022-08-24
Suggestion for easier access to xy values from PolyFile object


veenstrajelmer:
**Is your feature request related to a problem? Please describe.**
When plotting a pli file, currently I need to extract x/y locations with a for loop. Is it also possible to access the x/y coordinates directly or can this be added?
```
from hydrolib.core.io.polyfile.parser import read_polyfile #TODO: should be replaced with PolyFile after bugfix
file_pli = Path(r'c:\DATA\dfm_tools_testdata\GSHHS_high_min1000km2.ldb')
polyfile_object = read_polyfile(file_pli,has_z_values=False)
pli_PolyObjects = polyfile_object['objects']

fig,ax = plt.subplots()
for iPO, pli_PolyObject_sel in enumerate(pli_PolyObjects):
    print(f'processing PolyObject {iPO+1} of {len(pli_PolyObjects)}: name={pli_PolyObject_sel.metadata.name}')
    xvals = np.array([p.x for p in pli_PolyObject_sel.points])
    yvals = np.array([p.y for p in pli_PolyObject_sel.points])
    ax.plot(xvals,yvals)
```

**Describe the solution you'd like**
Access all xy values without looping over all points. Shapely for instance has a .coords.xy function:
https://stackoverflow.com/questions/20474549/extract-points-coordinates-from-a-polygon-in-shapely

veenstrajelmer:
This is also applicable for XYZModel and XYNmodel. Currently, I created these funtions, but it would be better if there would be some generic conversion tool: https://github.com/openearth/dfm_tools/blob/work_jelmer/dfm_tools/hydrolib_helpers.py

priscavdsluis:
Closing this issue as discussed with @veenstrajelmer.
It would be nice if we can find something more generic that will work for multiple objects. 
Jelmer can do that now easily locally.

I think if more users ask for this kind of functionality we can take a look at it in a more broader way.

PR #371, opened 2022-10-12
#329: Provide easier access to the x and y-coordinates of a PolyObject

priscavdsluis:
Also added a function for getting the x, y ánd z coordinates.


hydrolib/core/io/polyfile/models.py
```diff
@@ -76,6 +76,35 @@ class PolyObject(BaseModel):
     metadata: Metadata
     points: List[Point]
 
+    @property
+    def xy_coordinates(self) -> Tuple[List[float], List[float]]:
```
veenstrajelmer, line 80: Maybe xy_values/xyz_values instead of coordinates? Or is z also considered a coordinate?

hydrolib/core/io/polyfile/models.py
```diff
@@ -76,6 +76,35 @@ class PolyObject(BaseModel):
...
+        """Get a tuple containing the x-coordinates and y-coordinates of the polygon points.
+
+        Returns:
+            Tuple[List[float], List[float]]: Tuple containing a list of x-coordinates and a list of y-coordinates
+        """
+
+        x, y, z = self.xyz_coordinates
+        return x, y
```
veenstrajelmer, line 88: Maybe it would be nice to return it as a dict (or a pandas dataframe): {'x':x,'y':y}
In this case the user also knows what is inside. But maybe that would be unproper programming?

veenstrajelmer (changes requested):
This is now only a bound method of PolyObject. It would be valuable to also have it on XYZmodel, maybe XYNmodel and others. Furthermore, do you think it is also possible to add it to PolyFile or is that not nice programming?

hydrolib/core/io/polyfile/models.py
priscavdsluis: Good point I will update it to values.

hydrolib/core/io/polyfile/models.py
priscavdsluis: I think it is indeed not very clean to return a dictionary. We want to "shield" users from using strings to retrieve data, since that is error prone and requires that the user has knowledge about the key words. 

In this case, I would prefer returning a custom object over a dictionary such as:

```
class XYZValues:
    x: List[float] = []
    y: List[float] = []
    z: List[Optional[float]] = []

    def append(self, x: float, y: float, z: Optional[float]):
        self.x.append(x)
        self.y.append(y)
        self.z.append(z)
```

Which can be used as:
```
values = poly_object.xyz_coordinates
ax.plot(values.x, values.y)
```

As opposed to using a dict:
```
values = poly_object.xyz_coordinates
ax.plot(values["x"], values["y"])
```

Please let me know what you think.

veenstrajelmer: Also, if I try out your code, this does not fail:
```
xyz = XYZValues()
xyz.append(x=[1,2],y='a',z=[None,1])
```

hydrolib/core/io/polyfile/models.py
```diff
@@ -55,6 +55,31 @@ def _get_identifier(self, data: dict) -> Optional[str]:
...
+    x: List[float] = []
+    """The x values."""
+
+    y: List[float] = []
+    """The y values."""
+
+    z: List[Optional[float]] = []
+    """The z values. The values can be `None`."""
+
+    def append(self, x: float, y: float, z: Optional[float]):
```
veenstrajelmer, line 70: z: Optional[float]=None

veenstrajelmer (changes requested):
add default value for z

hydrolib/core/io/polyfile/models.py
priscavdsluis: I can also remove the append function on the XYZValues object and just set up the object correctly internally:

```
        values = XYZValues()

        for point in self.points:
            values.x.append(point.x)
            values.y.append(point.y)
            values.z.append(point.z)

        return values
```

I removed it in the last commit:
https://github.com/Deltares/HYDROLIB-core/commit/114cf2cd0f3434f9accabc249c6bc4afac8ff4f0

hydrolib/core/io/polyfile/models.py
priscavdsluis: `Optional` in this case means that it can be either a float or a `None`. Below the surface `Optional[float]` actually is a `Union[float, None]`. So it does not mean that this argument is optional for the user; they have to provide it, but it is allowed to be `None`
For more information: https://docs.python.org/3/library/typing.html#typing.Optional

However, I have removed this function (see comment above).

priscavdsluis:
> This is now only a bound method of PolyObject. It would be valuable to also have it on XYZmodel, maybe XYNmodel and possibly others.

@veenstrajelmer, thanks for the suggestion. I also added the functionality to the XYZModel.
We currently have no XYNModel.

priscavdsluis:
Closing this PR as discussed with @veenstrajelmer.
It would be nice if we can find something more generic that will work for multiple objects. 
Jelmer can do that now easily locally.

I think if more users ask for this kind of functionality we can take a look at it in a more broader way.

Issue #258, opened 2022-06-27
Create toplevel hydrolib.core.io.dflowfm tree structure for FMModel input


priscavdsluis:
**What is the need for this task.**
Currently in hydrolib.core.io many subdirectories exist for FM related modules. 
One subdirectory in hydrolib.core.io is meant for RR modules. 
hydrolib.core.io has grown a bit messy because of this. 

**What is the task?**
Move all FM related subdirecties in hydrolib.core.io to hydrolib.core.io.dflowfm

**Additional context**
Changing the namespace for all FM objects might require a lot of rework in the documentation and code that is dependent on it. 

See #168 for RR

priscavdsluis:
We should also consider updating and using the __init__.py files to make imports easier as advised by @BeardedPlatypus

tim-vd-aardweg:
Eventhough all the work for this issue has been done, we decided to wait until the next release to actually merge it, because this issue will obviously cause all sorts of merge conflicts.

It may be quicker to start this issue from scratch after the release, than to fix all the merge conflicts.

PR #392, opened 2022-10-28
Add toplevel hydrolib.core.io.dflowfm tree structure

tim-vd-aardweg:
Fixes: #258


tests/io/dflowfm/ini/test_ini.py
```diff
@@ -29,7 +29,7 @@
     write_ini,
 )
 
-from ..utils import test_output_dir
+from ....utils import test_output_dir
```
priscavdsluis, line 32: perfection 👌

hydrolib/core/io/dflowfm/bc/models.py
```diff
@@ -19,11 +19,15 @@
 from pydantic.fields import Field
 
 from hydrolib.core.basemodel import BaseModel
-from hydrolib.core.io.ini.io_models import Property, Section
-from hydrolib.core.io.ini.models import DataBlockINIBasedModel, INIGeneral, INIModel
```
priscavdsluis, line 23: General remark:
I am unsure about the INI. I think it is fine to leave it for now, since we only use in FM

priscavdsluis (approved):

Issue #385, opened 2022-10-25
Refactor the implementation of vectors in forcing models


tim-vd-aardweg:
In #316 support for vectors in forcing blocks with a function type of t3d and timeseries was added. However, after several discussions, we believe this may not be the best way to implement it.

We should have short discussion on how to improve the implementation.

I would suggest something like the following:
- Create a new `VectorForcingBase` model that inherits from `ForcingBase`
- Let `T3D` and `TimeSeries` inherit from this new base class.
- Change the `quantityunitpair` in the `ForcingBase` back to a list of `QuantityUnitPair`.
- Add a `quantityunitpair` property to the new `VectorForcingBase` that is a `List[ScalarOrVectorQUP]`
- Move the validation from the `VectorQUPValidation` into the `VectorForcingBase`.
- Update the testcase structure to reflect these changes.

PR #394, opened 2022-11-01
Refactor the support for vectors in .bc files.

tim-vd-aardweg:
Fixes: #385


hydrolib/core/io/dflowfm/bc/models.py
```diff
@@ -566,6 +593,20 @@ def _validate_quantityunitpairs(cls, values: Dict) -> Dict:
             else len(verticalpositions.split())
         )
 
+        return number_of_verticalpositions
+
+    @root_validator(pre=True)
+    def _validate_quantityunitpairs(cls, values: Dict) -> Dict:
+        super().validate_and_update_quantityunitpairs(values)
```
arthurvd, line 600: Isn't this now called twice? Because in `VectorForcingBase` is it also a `@root_validator`. Maybe check in debug mode with a breakpoint.

hydrolib/core/io/dflowfm/bc/models.py
```diff
@@ -617,12 +653,19 @@ def _validate_that_all_quantityunitpairs_have_valid_verticalpositionindex(
...
         verticalpositionindexes: List[int],
         number_of_verticalpositions: int,
-        quantityunitpairs: List[QuantityUnitPair],
+        quantityunitpairs: List[ScalarOrVectorQUP],
     ) -> None:
         if verticalpositionindexes is None:
             raise ValueError("vertPositionIndex is not provided")
 
-        if len(verticalpositionindexes) != len(quantityunitpairs) - 1:
+        all_qups = []
```
arthurvd, line 661: I find it not so elegant that we need to create a full list, only to map a function across the individual list elements below.
I understand the cause: contrary to the original implementation, now *first* the VectorQUPs get packed by the base class, and only after that the verticalpositionindexes are added by the T3D class.

This is comment is for brainstorming only: I'm still thinking if there is a more elegant way or not.

arthurvd (commented):
Let's do some more brainstorming this Thursday, please see my comments.

hydrolib/core/io/dflowfm/ini/util.py
```diff
@@ -261,6 +264,16 @@ def get_from_subclass_defaults(cls: Type[BaseModel], fieldname: str, value: str)
     """
     for c in cls.__subclasses__():
         default = c.__fields__.get(fieldname).default
+
+        # if default is none: zoek in de subclasses naar de default, recursively?
+        if default is None:
+            for sc in c.__subclasses__():
+                default = sc.__fields__.get(fieldname).default
```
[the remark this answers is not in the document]
tim-vd-aardweg: Added recursion to this function

hydrolib/core/io/dflowfm/bc/models.py
tim-vd-aardweg: Indeed! Let's say we create a `T3D`, then `ForcingBase._validate_quantityunitpair` is called, because it's a `root_validator` for the `ForcingBase`. Next, we call the `root_validator` in `VectorForcingBase`, where we call the same `ForcingBase._validate_quantityunitpair` again. Finally, the `root_validator` in the `T3D` class calls `VectorForcingBase.validate_and_update_quantityunitpairs` again, which in turn calls `ForcingBase._validate_quantityunitpair` a third time.

I will remove the calls to the base validators, as they are already called automatically.

Good catch! :)

tim-vd-aardweg: Done!

hydrolib/core/io/dflowfm/bc/models.py
arthurvd: Discussed this with @tim-vd-aardweg in a brief meeting. We decided:
* this new `VectorForcingBase` class is still an improvement, so we keep it.
* the above duplicate list `all_qups` is either acceptable, or we change it as follows:
  * pass the original list `quantityunitpairs` to the method `_add_verticalpositionindex_to_quantityunitpairs()` and in there: don't do a `zip`, but instead loop across `quantityunitpairs` and when encountering a `VectorQuantityUnitPairs`, do a nested loop over its children, and for these do the actual assignment of `.verticalpositionindex=...`.
  * while looping+nested-looping, maintain a counter of total nr of regular qups, and only at the end do a check wether that is equal to the length of the input `verticalpositionindexes` list.

hydrolib/core/io/dflowfm/ini/util.py
```diff
@@ -260,13 +263,29 @@ def get_from_subclass_defaults(cls: Type[BaseModel], fieldname: str, value: str)
...
-        default = c.__fields__.get(fieldname).default
-        if default.lower() == value.lower():
+        default = _try_get_default_value(c, fieldname, value)
+        if default is not None and default.lower() == value.lower():
             return default
-
     return value
 
 
+def _try_get_default_value(c, fieldname, value):
```
arthurvd, line 272: Consider removing this method and instead do all in `get_from_subclass_defaults()` with these modifications:
```python
def get_from_subclass_defaults(cls: Type[BaseModel], fieldname: str, value: str, return_value_if_not_found: bool = True) -> str:
    """
    Gets a value that corresponds with the default field value of one of the subclasses.
    If the subclass doesn't have the specified field, it will look into its own subclasses
    recursively for the specified fieldname.

    Args:
        cls (Type[BaseModel]): The parent model type.
        fieldname (str): The field name for which retrieve the default for.
        value (str): The value to compare with.

    Returns:
        [type]: The field default that corresponds to the value.
    """
    for c in cls.__subclasses__():
        default = get_from_subclass_default(c, fieldname, value, False )
        if default is not None and default.lower() == value.lower():
            return default
    if return_value_if_not_found:
        return value
    else:
        return None
```

hydrolib/core/io/dflowfm/bc/models.py
```diff
@@ -423,7 +281,186 @@ def __repr__(self) -> str:
         return str(representable)
 
 
-class TimeSeries(ForcingBase):
+class VectorForcingBase(ForcingBase):
```
arthurvd, line 284: @tim-vd-aardweg: I just mentioned to consider renaming to `ScalarOrVectorForcingBase`, but forget about that, the name is fine.

arthurvd (changes requested):
@tim-vd-aardweg : See the remaining comments after our meeting of this afternoon.

hydrolib/core/io/dflowfm/bc/models.py
tim-vd-aardweg: I removed the creation of a separate list of qups. We now just iterate over all `ScalarOrVectorQUP` and unpack any `VectorQuantityUnitPairs` we find.

hydrolib/core/io/dflowfm/bc/models.py
tim-vd-aardweg: Okay :)

hydrolib/core/io/dflowfm/ini/util.py
tim-vd-aardweg: This won't work :(

Let's assume we call this function in the context of `ForcingBase` and we want to see if any of its subclasses has a default value of `harmonic` for the `function` property.

In the `for` loop, we iterate over the children of `ForcingBase`. At some point we reach the point where `c` is equal to the `Harmonic` class. The first line in the `for` loop recursively calls the `get_from_subclass_defaults` function. So we loop over the children of the `Harmonic` class. But there are none. So the call to `get_from_subclass_defaults` return None and we eventually leave this function without returning the correct default value.

So what we would need above the `for` loop, is to check if the current class has a `fieldname` property with a default value of `value`. However, we have to change the name of the function, as it no longer only check children, but it also checks the base class. Furthermore, it will add code duplication again. That said, I am starting to think my previous implementation may still be a good candidate.

hydrolib/core/io/dflowfm/bc/models.py
```diff
@@ -389,12 +239,20 @@ def validate(cls, v):
         # https://github.com/samuelcolvin/pydantic/pull/2336
         if isinstance(v, dict):
             for c in cls.__subclasses__():
```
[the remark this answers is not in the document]
tim-vd-aardweg: Created a utility function that now fully recursively looks in the children and their children, etc.

hydrolib/core/io/dflowfm/ini/util.py
arthurvd: We discussed this Monday, and indeed we need this. I have committed some final improvements in 0e24282.

arthurvd (approved):

PR #401, opened 2022-11-07
#355: Add support for non-recursively loading models

### LINKED ISSUE #355: Make it possible to choose whether you want to load a model recursively or not.
Some models take very long to load, while there is no need to deserialize the whole model. Some users just want to change a small thing in one of the models and save it. It would be handy if there was an option to control if the model should recursively be loaded or not.

Add an option when loading the model: `recurse`, like we have with the `save` function, e.g.:
`model.save(recurse=False)`

priscavdsluis:
Fixes #355

### DIFF
--- a/hydrolib/core/basemodel.py
+++ b/hydrolib/core/basemodel.py
@@ -440,6 +440,14 @@ class FileModelCache:
         """
         self._cache_dict[path] = model
 
+    def is_empty(self) -> bool:
+        """ Whether or not this file model cache is empty.
+
+        Returns:
+            bool: Whether or not the cache is empty.
+        """
+        return not any(self._cache_dict)
+
 
 class FileLoadContext:
     """FileLoadContext provides the context necessary to resolve paths
@@ -452,6 +460,8 @@ class FileLoadContext:
         self._path_resolver = FilePathResolver()
         self._cache = FileModelCache()
         self._file_casing_resolver = FileCasingResolver()
+        self._recurse_initialized = False
+        self._recurse = False
 
     def retrieve_model(self, path: Optional[Path]) -> Optional["FileModel"]:
         """Retrieve the model associated with the path.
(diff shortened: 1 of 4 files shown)


hydrolib/core/basemodel.py
```diff
@@ -646,13 +696,20 @@ def __init__(
         Args:
             filepath (Optional[Path], optional): The file path. Defaults to None.
             resolve_casing (bool, optional): Whether or not to resolve the file name references so that they match the case with what is on disk. Defaults to False.
+            recurse (bool): Whether or not to recursively load the model. Default to True.
```
tim-vd-aardweg, line 699:
```suggestion
            recurse (bool, optional): Whether or not to recursively load the model. Defaults to True.
```

hydrolib/core/basemodel.py
```diff
@@ -646,13 +696,20 @@ def __init__(
...
         """
         if not filepath:
             super().__init__(*args, **kwargs)
             return
 
         with file_load_context() as context:
-            context.set_resolve_casing(resolve_casing)
+            context.initialize_load_settings(recurse, resolve_casing)
+
+            if not context.load_settings.recurse and not context.cache_is_empty():
```
tim-vd-aardweg: Perhaps we can add a comment or put this in a separate function as it may not be clear at first glance when we should go into this if-statement.

hydrolib/core/basemodel.py
```diff
@@ -734,6 +734,17 @@ def __init__(
 
             context.pop_last_parent()
 
+    @classmethod
+    def _should_load_model(cls, context: FileLoadContext) -> bool:
```
tim-vd-aardweg, line 738: Nice! :)

tim-vd-aardweg (approved):
Review OK!

Issue #399, opened 2022-11-03
LocationType is written to file as the full type e.g. "LocationType.oned"


priscavdsluis:
**Describe the bug**
The full enum value type is written to file for LocationType. Maybe there are other enums as well where this happens.

**To Reproduce**
```
from hydrolib.core.io.crosssection.models import CrossLocModel
crossloc_model = CrossLocModel(filepath="crsloc.txt")
crossloc_model.save() 
```

If you inspect the file you will see that the cross sections are written as:
```
[CrossSection]
    id           = 99
    branchId     = 9
    chainage     = 403.089709
    shift        = 0.0
    definitionId = 99
    locationtype = LocationType.oned
```

**Expected behavior**
That the LocationType are written with their string value, in this case "1d".

**Cross section location file**
[crsloc.txt](https://github.com/Deltares/HYDROLIB-core/files/9926641/crsloc.txt)

**Version info (please complete the following information):**
 - Version: 0.3.1

tim-vd-aardweg:
This indeed happens with (all?) enums. For example in the `T3D` class, we have several attributes that are enums. For the default value we had to explicitly say something like:

`timeinterpolation: TimeInterpolation = Field(
        TimeInterpolation.linear.value, alias="timeInterpolation"
    )`

Otherwise it would write `TimeInterpolation.Linear` to the file instead of `linear`.

xldeltares:
Hi @priscavdsluis , I encountered this issue while reading the file back to hydrolib-core object.  I wonder if the reader needs some changes as well?'

**To Reproduce**
read the following crsloc.ini using 
```
CrossLocModel(filepath =Path( r"crsloc.ini"))
```
```
[General]
    fileVersion = 3.00
    fileType    = crossLoc

[CrossSection]
    id           = river_1_459.86 # Unique cross-section location id.
    branchId     = river_1        # Branch on which the cross section is located.
    chainage     = 459.86         # Chainage on the branch (m).
    shift        = -10.0          # Vertical shift of the cross section definition [m]. Defined positive upwards.
    definitionId = 1              # Id of cross section definition.
    locationtype = 1d

[CrossSection]
    id           = river_1_797.90 # Unique cross-section location id.
    branchId     = river_1        # Branch on which the cross section is located.
    chainage     = 797.899        # Chainage on the branch (m).
    shift        = 0.0            # Vertical shift of the cross section definition [m]. Defined positive upwards.
[...]

PR #403, opened 2022-11-07
Enums are incorrectly written to files

tim-vd-aardweg:
Fixes: #399

### DIFF
--- a/hydrolib/core/io/dflowfm/ini/models.py
+++ b/hydrolib/core/io/dflowfm/ini/models.py
@@ -1,5 +1,6 @@
 import logging
 from abc import ABC
+from enum import Enum
 from typing import Any, Callable, List, Literal, Optional, Set, Type, Union
 
 from pydantic import Extra, Field, root_validator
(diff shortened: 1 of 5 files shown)


hydrolib/core/io/dflowfm/ini/models.py
```diff
@@ -129,6 +130,8 @@ def _convert_value(cls, key: str, v: Any) -> str:
             return str(int(v))
         elif isinstance(v, list):
             return cls.get_list_field_delimiter(key).join([str(x) for x in v])
+        elif isinstance(v, Enum):
```
priscavdsluis, line 133: Nice! So this means that the `.value` is no longer necessary?

priscavdsluis (approved):

hydrolib/core/io/dflowfm/ini/models.py
tim-vd-aardweg: Indeed! I did not yet do a search through all the code to remove all occurances of the .value though.

PR #413, opened 2022-11-09
412: Extend build basic model notebook as post-install test notebook

### LINKED ISSUE #412: Create installation test notebook.
**What is the need for this task.**
When new users are beginning to use hydrolib-core, there's several ways to do the installation.
In order to test whether installation was successful and all of hydrolib-core is available, it would be nice to have a small test notebook that checks some basis hydrolib.core imports and shows the user the result.

**What is the task?**
* Re-use existing notebook `docs/tutorials/build_basic_model.ipynb`
* Add some message to confirm succesful package import
* Add some output to show that model files were indeed saved.

arthurvd:
Fixes #412.

### DIFF
--- a/docs/tutorials/build_basic_model.ipynb
+++ b/docs/tutorials/build_basic_model.ipynb
@@ -10,17 +10,38 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Import HYDROLIB-core functionalities and setup D-Flow FM model\n",
+    "The commands below import all needed packages and will prepare a model that will be saved in a subdir `testmodel/` within the current working dir."
+   ]
+  },
+  {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


docs/tutorials/build_basic_model.ipynb
```diff
@@ -9,18 +9,39 @@
     "This _Build a basic model_ tutorial illustrates how to build a simple D-Flow FM model from scratch using [HYDROLIB-core](https://github.com/Deltares/HYDROLIB-core). "
    ]
```
priscavdsluis, line 10: @arthurvd, I think the notebook suffices for testing whether hydrolib-core installation has succeeded. However, the imports that we use here are from an unreleased version. I assume that users will install the latest released version (0.3.1) and will not use the main. The main has the new `dflowm` package, these imports will not work with 0.3.1.

For my demo I have created a branch from release 0.3.1. You might consider doing that as well.

arthurvd: You are right. I did that, and created a draft PR for the 0.3.1-compatible version.
I'll close this PR now without merging.

Issue #425, opened 2022-11-18
Improve performance writing BC/INI files


priscavdsluis:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem or missing functionality is.

**Describe the solution you'd like**

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you've considered (if applicable).

**Additional context**

PR #426, opened 2022-11-18
#425: Improve performance of INI/BC serialization

priscavdsluis:


priscavdsluis:
With these changes the @veenstrajelmer tests will be about 33% faster (from ~30s to ~20s on 100000 rows)

Issue #450, opened 2023-03-01
Update MeshKernel to 2.0.2


priscavdsluis:
MeshKernel 2.0.2 compared to 2.0.0 now also supports Linux

**TODO**
- Update MeshKernelPy to 2.0.2

PR #453, opened 2023-03-07
#450: Update MeshKernel to 2.0.1

priscavdsluis:


priscavdsluis:
Will be replaced with #468 (MeshKernel 2.0.2)

PR #458, opened 2023-03-10
Fix: XYZ parser should support whitespace in lines and between lines

### LINKED ISSUE #415: xyz file with initial whitespace on lines is not correctly parsed by XYZmodel
**Describe the bug**
Initial whitespaces on each line in an xyz-file cause validation error:
```
   37115.2500000  419886.6875000     500.0000000
```

Reading a file like this, results in an error:
```
diffusivity_rivzee_v6.xyz -> points -> 1168853 -> x: y:37115.2500000 z:419886.6875000 -> x
  value is not a valid float (type=type_error.float)
```
x column is seen as y column etc.

**To Reproduce**
```
from pathlib import Path
try: #0.3.1 release
    from hydrolib.core.io.xyz.models import XYZModel
except: #main branch and next release #TODO: move to easy imports after https://github.com/Deltares/HYDROLIB-core/issues/410
    from hydrolib.core.io.dflowfm.xyz.models import XYZModel
#file_xyz = r'p:\11205258-006-kpp2020_rmm-g6\C_Work\08_RMM_FMmodel\geometry_j19_NL_6-v2\rmm_vzm_v1p1_initial_water_level.xyz'
file_xyz = r'p:\11206813-006-kpp2021_rmm-2d\C_Work\31_RMM_FMmodel\general\diffusivity_rivzee_v6.xyz'
data_xyz = XYZModel(Path(file_xyz))
```

 - OS: Windows and Spyder
 - Version hydrolib main branch

arthurvd:
Fixes #415


hydrolib/core/dflowfm/xyz/parser.py
```diff
@@ -22,13 +22,20 @@ def parse(filepath: Path) -> Dict:
...
 
-                if len(line) < 5:  # 3 values, two whitespaces
+                line = line.strip()
+                if line.startswith("*") or len(line) == 0:
                     continue
 
-                x, y, z, c = re.split(xyzpattern, line, maxsplit=3)
+                try:
+                    x, y, z, *c = re.split(xyzpattern, line, maxsplit=3)
+                except ValueError:
```
tim-vd-aardweg, line 33: Can you add documentation to this method to indicate that this method may throw a `ValueError`?

hydrolib/core/dflowfm/xyz/parser.py
```diff
@@ -22,13 +22,20 @@ def parse(filepath: Path) -> Dict:
...
+            for linenr, line in enumerate(f.readlines()):
 
-                if len(line) < 5:  # 3 values, two whitespaces
+                line = line.strip()
+                if line.startswith("*") or len(line) == 0:
                     continue
 
-                x, y, z, c = re.split(xyzpattern, line, maxsplit=3)
+                try:
+                    x, y, z, *c = re.split(xyzpattern, line, maxsplit=3)
```
tim-vd-aardweg, line 32: Is it required to use `*c` (instead of simply `c`) since we specify `maxsplit=3`? So `c` will contain at most one string?

hydrolib/core/dflowfm/xyz/parser.py
```diff
@@ -22,13 +22,20 @@ def parse(filepath: Path) -> Dict:
...
+                except ValueError:
+                    raise ValueError(
+                        f"Error parsing XYZ file '{filepath}', line {linenr+1}."
```
tim-vd-aardweg, line 35: Nice! Good catch that you should increase the `linenr` by 1! 👍

tests/dflowfm/test_xyz.py
```diff
@@ -1,7 +1,17 @@
+from pathlib import Path
```
tim-vd-aardweg, line 1: Not used, can be removed

tests/dflowfm/test_xyz.py
```diff
@@ -34,8 +44,44 @@ def test_save(self):
         ]
         model = XYZModel(points=points)
 
+        if output_file.is_file():
```
tim-vd-aardweg, line 47: What are we doing here? Is this something I should be doing as well in my tests?

hydrolib/core/dflowfm/xyz/parser.py
arthurvd: Since the comment (c) is optional, the split might not return a fourth value, then incorrectly raising a ValueError.
When using *c, the fourth value will be a list that may be *empty*, which nicely handles optional comments.

tests/dflowfm/test_xyz.py
arthurvd: Done in new commit.

hydrolib/core/dflowfm/xyz/parser.py
arthurvd: Done in new commit.

tests/dflowfm/test_xyz.py
arthurvd: It deletes the file if it already exists. That makes sure that we are not getting a green testcase that is accidentally caused by an older file still being present. I borrowed it from old code that used to be present in `test_model.py`.

tim-vd-aardweg: Thanks! 👍

tim-vd-aardweg (approved):
Review OK!

tim-vd-aardweg:
There is 1 code smell remaining, but I suggest we ignore it. My preference is to keep the name of the test file in the test itself and not as a shared variable somewhere.

PR #465, opened 2023-03-13
feat: Support .tim files

MRVermeulenDeltares:
Refs: #348


hydrolib/core/dflowfm/tim/models.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import Callable, Dict, List
+
+from hydrolib.core.basemodel import ParsableFileModel, SerializerConfig
+
+from .parser import TimParser
+from .serializer import TimSerializer, TimSerializerConfig, TimTimeSerie
+
+
+class TimModel(ParsableFileModel):
+    """Model or tim data.
```
priscavdsluis, line 11:
```suggestion
    """Class representing a tim (*.tim) file.
```

hydrolib/core/dflowfm/tim/parser.py
```diff
@@ -0,0 +1,91 @@
...
+        self.time = time
+        self.series = series
+        self.comment = comment
+
+
+class TimParser:
+    """
+    A parser for .tim files.
+    Full line comments are supported.
+    Partial comments will have the line skipped.
```
priscavdsluis, line 23: So if you have a comment after your values the whole line will be skipped?

hydrolib/core/dflowfm/tim/models.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Attributes:
```
priscavdsluis, line 13: Document the attributes

hydrolib/core/dflowfm/tim/parser.py
```diff
@@ -0,0 +1,91 @@
...
+from pathlib import Path
+from typing import List
+
+timpattern = re.compile(r"\s+")
+
+
+class TimTimeSerie:
+    comment: str
+    time: float
+    serie: List[float]
```
priscavdsluis, line 11:
```suggestion
    series: List[float]
```

hydrolib/core/dflowfm/tim/parser.py
```diff
@@ -0,0 +1,91 @@
+import re
+from pathlib import Path
+from typing import List
+
+timpattern = re.compile(r"\s+")
+
+
+class TimTimeSerie:
```
priscavdsluis, line 8: Serie is not the singular form of series. Maybe rename it to `TimTimeData` or `TimTimeSeriesRow` ?

hydrolib/core/dflowfm/tim/parser.py
```diff
@@ -0,0 +1,91 @@
...
+            filepath (Path): .tim file to be read.
+
+        Returns:
+            List: List of "TimTimeSerie".\n
+            When file is empty returns an empty list.
+        """
+        timeseries = []
+        with filepath.open() as file:
+            for line in file.readlines():
+                TimParser._read_line(line, timeseries)
```
priscavdsluis, line 40: I would not add the time series within this method.
I would put the none check here instead of (`_add_valid_timeserie`)

hydrolib/core/dflowfm/tim/parser.py
```diff
@@ -0,0 +1,91 @@
...
+        timeseries.append(timeserie)
+
+    @staticmethod
+    def _create_timeserie(time: str, series: List[str]):
+        if TimParser._not_numeric(time):
+            return
+
+        listofvalues = []
+        for value in series:
+            if TimParser._not_numeric(value):
```
priscavdsluis, line 74: So this line is skipped if it contains a comment? 
I don't think that is the correct functionality, also I see the following information in the issue:

> Comments moeten gelezen en geschreven kunnen worden
> - de comments moeten ingelezen worden en ook worden terug gezet als er wordt geschreven

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @staticmethod
+    def _write_row(
+        file,
+        serializer: TimTimeSeriesSerializer,
+        timeserie: TimTimeSerie,
+        config: TimSerializerConfig,
+    ):
+        if timeserie.comment:
+            file.write(timeserie.comment)
```
priscavdsluis, line 97: The serializer is responsible for putting the `#`/`*` in front of a comment in my opinion. User should be able to set the comment to `blah` and the serializer should write `# blah` in the file.

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
...
+from hydrolib.core.dflowfm.ini.serializer import MaxLengths, SectionSerializer
+from hydrolib.core.dflowfm.tim.parser import TimTimeSerie
+
+
+class TimSerializerConfig(SerializerConfig):
+    datablock_indent: int = 4
+    datablock_spacing: int = 2
+
+
+class TimTimeSeriesSerializer(SectionSerializer):
```
priscavdsluis, line 15: I am not sure if inheriting from the `SectionSerializer` would be a good idea since that is a INI specialized class.
In my opinion, it is fine if you use the stuff that you need from the `SectionSerializer` and put it here, because INI and TIM are just not the same file formats and thus the same serializer should not be used. 

Maybe we can look it this together.

hydrolib/core/dflowfm/tim/models.py
```diff
@@ -0,0 +1,39 @@
...
+        timeseries: List[TimTimeSerie]
+        serializer_config: TimSerializerConfig
+    """
+
+    serializer_config = TimSerializerConfig()
+    timeseries: List[TimTimeSerie]
```
priscavdsluis, line 19: Was the following point addressed? 

> error bij meerdere keren dezelfde tijd

hydrolib/core/dflowfm/tim/models.py
```diff
@@ -0,0 +1,39 @@
...
+
+from .parser import TimParser
+from .serializer import TimSerializer, TimSerializerConfig, TimTimeSerie
+
+
+class TimModel(ParsableFileModel):
+    """Model or tim data.
+
+    Attributes:
+        timeseries: List[TimTimeSerie]
```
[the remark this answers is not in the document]
MRVermeulenDeltares: I did that as first, but what should I do if a line is a comment?

hydrolib/core/dflowfm/tim/parser.py
```diff
@@ -0,0 +1,91 @@
...
+                return
+            listofvalues.append(float(value))
+
+        return TimTimeSerie(time=float(time), series=listofvalues)
+
+    @staticmethod
+    def _not_numeric(value: str):
+        return not value.replace(",", "").replace(".", "").replace("-", "").isnumeric()
```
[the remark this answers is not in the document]
MRVermeulenDeltares: non numeric values in a line are not supported: "Do NOT support comments at the end of regular data lines"
I added this for skipping the line, would raising an error be better in this case?

hydrolib/core/dflowfm/tim/models.py
MRVermeulenDeltares: This was discussed when the List was a dictionary, lets discuss further

hydrolib/core/dflowfm/tim/parser.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        timeseries = []
+        with filepath.open() as file:
+            for line in file.readlines():
+                TimParser._read_line(line, timeseries)
+        return timeseries
+
+    @staticmethod
+    def _read_line(line, timeseries: List[TimTimeSerie]):
+        if TimParser._line_is_comment(line):
```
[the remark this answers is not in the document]
MRVermeulenDeltares: Lets further discuss, also in regard to the serializer

hydrolib/core/dflowfm/tim/parser.py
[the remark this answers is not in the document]
MRVermeulenDeltares: done

hydrolib/core/dflowfm/tim/__init__.py
tim-vd-aardweg, line 1: Also, do not forget to add the following contents to this `__init__.py` file:

```
from .models import TimModel

__all__ = [
    "TimModel",
]
```

And since you have created a new directory (`hydrolib/core/dflowfm/tim`), you also have to update the `hydrolib\core\dflowfm\__init__.py` file by adding:

`from .tim import *`

This allows user to do the following:

```
import hydrolib.core.dflowfm as hcdfm
hcdfm.TimModel(<path>)
```

Every time you create a new class that the users can/should interact with, we have to explicitly add them to the `__init__.py` files to expose them to the user. (They are already exposed ofcourse, but adding them to the `__init__.py` allows them to use the classes via a shortcut / cleaner way.)

tim-vd-aardweg (changes requested):

hydrolib/core/dflowfm/tim/__init__.py
MRVermeulenDeltares: done!
Thx for the reminder!

priscavdsluis:
> The 22.0% Duplication is located in the test, how should I handle this?

As discussed with @MRVermeulenDeltares, we can ignore duplication in tests.

hydrolib/core/dflowfm/tim/models.py
```diff
@@ -0,0 +1,42 @@
...
+from typing import Callable, Dict
+
+from hydrolib.core.basemodel import ModelSaveSettings, ParsableFileModel
+
+from .parser import TimParser
+from .serializer import TimSerializer, TimSerializerConfig
+
+
+class TimModel(ParsableFileModel):
+    """Class representing a tim (*.tim) file.
```
priscavdsluis, line 11:
```suggestion
    """Class representing a *.tim file.
```

hydrolib/core/dflowfm/tim/models.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Attributes:
+        data: Dictionary with keys \"comment\" & time as numeric and value as List of floats".\n
+        serializer_config: TimSerializerConfig
+    """
+
+    serializer_config = TimSerializerConfig()
+
+    timeseries: Dict
```
priscavdsluis, line 20:
```suggestion
    timeseries: Dict[float, List[float]]
```

hydrolib/core/dflowfm/tim/parser.py
```diff
@@ -0,0 +1,75 @@
...
+from pathlib import Path
+from typing import Dict
+
+timpattern = re.compile(r"\s+")
+
+
+class TimParser:
+    """
+    A parser for .tim files.
+    Full line comments at the start of the file are supported by * and #.
```
priscavdsluis, line 11:
```suggestion
    Full line comments at the start of the file are supported. Comment lines start with either a `*` or a `#`.
```

hydrolib/core/dflowfm/tim/parser.py
```diff
@@ -0,0 +1,75 @@
+import re
```
priscavdsluis, line 1: I have looked at this class, and I think we could simplify some of the code. 

Assumptions for a correct file:
- Comments can only be present before the time series data
- Time series data are rows of data delimited by spaces
- Time series data cannot contain duplicate times

The `TimModel` is responsible for validating the actual data:
- The time series data should all be floating values (= strings can be converted to floats successfully)
- The time series data should have enough data. 

With these rules you could make a pretty simple parser.
- The parser does not have to check for correct floating values or convert them.
- When dealing with comments, only the first "#" or "*" should be stripped off. I think it would be unexpected behaviour if all the leading characters are stripped off.

Please take a look at the following code and tell me what your opinion is.

"""
from pathlib import Path
from typing import Any, Dict, List

class TimParser:
    """A parser for .tim files that extracts comments and time series data."""

    @staticmethod
    def parse(filepath: Path) -> Dict[str, Any]:
        """Parse a .tim file into a dictionary with comments and time series data.

        Args:
            filepath (Path): Path to the .tim file to be read.

        Returns:
            Dict[str, Any]: A dictionary with keys "comments" and "timeseries", where "comments"
                  is a list of strings representing comments found at the start of the file, and
                  "timeseries" is a dictionary where each key is a time and each value
                  is a list of strings.

        Raises:
            ValueError: If the file contains a comment that is not at the start of the file or
             if the time series contains a duplicate time entry.
        """

        comments: List[str] = []
        timeseries: Dict[str, List[str]] = {}

        with filepath.open() as file:
            lines = file.readlines()

            # Read header comments
            start_timeseries_index = 0
            for line_index in range(len(lines)):

                line = lines[line_index].strip()

                if len(line) == 0:
                    comments.append(line)
                    continue
                
                if line.startswith("#") or line.startswith("*"):
                    comments.append(line[1:])
                    continue

                start_timeseries_index = line_index
                break

            # Read time series data
            for line_index in range(start_timeseries_index, len(lines)):
                line = lines[line_index].strip()

                if len(line) == 0:
                    continue

                TimParser._raise_error_if_contains_comment(line, line_index + 1)

                time, *values = line.split()

                TimParser._raise_error_if_duplicate_time(time, timeseries, line_index + 1)

                timeseries[time] = values

        return {"comments" : comments, "timeseries" : timeseries}

    @staticmethod
    def _raise_error_if_contains_comment(line: str, line_index: int):
        if "#" in line or "*" in line:
            raise ValueError(f"Line {line_index}: comments are only supported at the start of the file, before the time series data.")
    
    @staticmethod
    def _raise_error_if_duplicate_time(time: str, timeseries: Dict[str, List[str]], line_index: int):
        if time in timeseries:
            raise ValueError(f"Line {line_index}: time series cannot contain duplicate times. Time: {time}")
"""

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
+from pathlib import Path
+from typing import Dict, Iterable, Optional
+
+from hydrolib.core.basemodel import ModelSaveSettings, SerializerConfig
+from hydrolib.core.dflowfm.ini.io_models import Datablock, DatablockRow
```
priscavdsluis, line 5: Don't use INI

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @staticmethod
+    def _serialize_row_element(elem: str, config) -> str:
+        max_length = config.max_length_datablock  # type: ignore
+        whitespace = _get_offset_whitespace(elem, max_length)
+        return elem + whitespace
+
+
+def _get_offset_whitespace(key: Optional[str], max_length: int) -> str:
+    key_length = len(key) if key is not None else 0
```
priscavdsluis, line 99: Can key be None?

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
+from pathlib import Path
+from typing import Dict, Iterable, Optional
+
+from hydrolib.core.basemodel import ModelSaveSettings, SerializerConfig
+from hydrolib.core.dflowfm.ini.io_models import Datablock, DatablockRow
+
+
+class TimSerializerConfig(SerializerConfig):
+    max_length_datablock = 10
```
priscavdsluis, line 9: Why is this an attribute? The max length should be derived, it should not be set with a constant value I think.

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
...
+            config (TimSerializerConfig): The serialization configuration.
+            save_settings (ModelSaveSettings): The model save settings.
+        """
+        path.parent.mkdir(parents=True, exist_ok=True)
+
+        datarowlist = []
+        commentrowlist = []
+
+        for key in data:
+            if key == "comments":
```
priscavdsluis, line 36: Why not just get the data that you want? 
with `data["comments"] and data["timeseries"]`

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
...
+        data: Dict,
+        config: TimSerializerConfig,
+        save_settings: ModelSaveSettings,
+    ) -> None:
+        """
+        Serializes the timeseries data to the file at the specified path in .tim format.
+
+        Attributes:
+            path (Path): The path to the destination file.
+            data (Dict): The data to be serialized.
```
priscavdsluis, line 26: Specify the expected data (which keys, which types, etc)

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
...
+            linesofdatablock.append(TimSerializer._serialize_row(row, config))
+        return linesofdatablock
+
+    @staticmethod
+    def _serialize_row(row: DatablockRow, config) -> str:
+        elem_spacing = " " * config.datablock_spacing
+
+        elems = []
+        for item in row:
+            if isinstance(item, str):
```
priscavdsluis, line 87: Can time series contain strings?

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
...
+
+        timeseries = TimSerializer._serialize_datablock(datarowlist, config)
+
+        with path.open("w") as file:
+            TimSerializer._write_rows(file, commentrowlist)
+            TimSerializer._write_rows(file, timeseries)
+
+    @staticmethod
+    def _write_rows(file, rowlist):
+        for rows in rowlist:
```
priscavdsluis, line 49: This is not very readable: `rowlist`, `rows`, `row`. Hard to know what is what.

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
...
+        rowlist = [str]
+        for comment in dictionary[key]:
+            if isinstance(comment, str):
+                data = f"# {comment}"
+                rowlist.append(data)
+        return rowlist
+
+    @staticmethod
+    def _serialize_datablock(datablock: Optional[Datablock], config) -> Iterable[str]:
+        if datablock is None or config.max_length_datablock is None:
```
priscavdsluis, line 73: Can datablock be none?

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
...
+        for value in dictionary[key]:
+            rowlist.append(str(value))
+        return rowlist
+
+    @staticmethod
+    def _get_commentblockrow(key, dictionary):
+        rowlist = [str]
+        for comment in dictionary[key]:
+            if isinstance(comment, str):
+                data = f"# {comment}"
```
priscavdsluis, line 67: There should be no space between # and the comments.
If you read - write - read - write, etc, there will be an extra space each time you write the file.

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,100 @@
+from pathlib import Path
+from typing import Dict, Iterable, Optional
```
priscavdsluis, line 2: Like the parser I think this serializer contains too much logic, for example some None checks or isinstance checks, which I believe are unnecessary. I really advise to split the `comments` and the `timeseries` attributes on the TimModel. This will make it easier to parse/serialize. 

Further more the logic with the column length is a little hard to read.

I thought about how to improve the readability and came up with the following:

```python
from pathlib import Path
from typing import Any, Dict, List

from hydrolib.core.basemodel import ModelSaveSettings, SerializerConfig

TimeSeriesRow = List[str]
TimeSeriesBlock = List[TimeSeriesRow]

class TimSerializerConfig(SerializerConfig):
    """Configuration settings for the TimSerializer."""

    column_spacing: int = 1
    """(int): The number of spaces to include between columns in the serialized .tim file."""

class TimSerializer:

    @staticmethod
    def serialize(
        path: Path,
        data: Dict[str, Any],
        config: TimSerializerConfig,
        save_settings: ModelSaveSettings,
    ) -> None:
        """
        Serializes timeseries data to a file in .tim format.

        Args:
            path (Path): The path to the destination .tim file.
            data (Dict[str, Any]): The timeseries data to be serialized. The data should be provided as a dictionary with the following keys:
                - 'comments' (List[str]): A list of comments to be included at the beginning of the file. Each comment should be a string.
                - 'timeseries' (Dict[float, List[float]]): A dictionary representing the timeseries data, where the key is the time and the value the values at that time.
            config (TimSerializerConfig): The serialization configuration settings.
            save_settings (ModelSaveSettings): The save settings to be used.
        """
        path.parent.mkdir(parents=True, exist_ok=True)

        lines = []

        # Serialize comments
        for comment in data["comments"]:
            lines.append(f"#{comment}")

        # Convert time series float data to time series string data        
        format_float = lambda v: f"{v:{config.float_format}}"

        timeseries_block: TimeSeriesBlock = []
        for time, row_elements in data["timeseries"].items():           
            timeseries_row = [format_float(time)] + [format_float(value) for value in row_elements] 
            timeseries_block.append(timeseries_row)
        
        # Make sure the columns are aligned and have the proper spacing
        column_space = " " * config.column_spacing
        max_length_columns = TimSerializer._get_column_lengths(timeseries_block)
        
        for timeseries_row in timeseries_block:

            row_elements: List[str] = []
            for index, value in enumerate(timeseries_row):
                whitespace_offset = " " * max_length_columns[index]
                row_elements.append(value + whitespace_offset)

            line = column_space.join(row_elements)
            lines.append(line)

        # Write all the line to file
        file_content = "\n".join(lines)
        with path.open("w") as file:
            file.write(file_content)

    @staticmethod
    def _get_column_lengths(timeseries_block: TimeSeriesBlock) -> List[int]:
        if len(timeseries_block) == 0:
            return []

        n_columns = len(timeseries_block[0])
        column_lengths = [0] * n_columns

        for timeseries_row in timeseries_block:
            for index, row_element in enumerate(timeseries_row):
                if len(row_element) > column_lengths[index]:
                    column_lengths[index] = len(row_element)

        return column_lengths

```

This serializer first serializes the comments leading with a # and adds it to the lines to writes.
Then it converts all the time series data (floats) to time series data with formatted floats (strings).
From this data the column lengths are derived.
Then the column spacing is used to space the columns. 

Then the file content is written in the end. 

Of course some parts could be extracted, but I think the flow and the order of operations is more clear.

Please let me know what you think.

hydrolib/core/dflowfm/tim/serializer.py
priscavdsluis: Please correct me if I am wrong.
I just think the behaviour will be very weird when some values are longer than 10 chars.

hydrolib/core/dflowfm/tim/parser.py
MRVermeulenDeltares: I tried to implement it in a similar fashion, but with splitting the comment and timeseries further up, it makes it a lot clearer!

Might it be an idea to also move the Read header comments and Read time series data into "private" methods?
Then we would get something like:

with filepath.open() as file:
            lines = file.readlines()
            comments = readcomment(lines)
            timeseries = readtimeseries(lines)
return {"comments" : comments, "timeseries" : timeseries}

MRVermeulenDeltares: Should we skip if the line contains no or too little information?

MRVermeulenDeltares: What about when a line contains no numeric value? or should this be handled at timmodel level?

MRVermeulenDeltares: What I understood is that errors are preferred over warnings

MRVermeulenDeltares: Do we also want to put the file location/name in the error messages?

MRVermeulenDeltares: I will also add location to the error message

hydrolib/core/dflowfm/tim/models.py
```diff
@@ -0,0 +1,73 @@
...
+class TimModel(ParsableFileModel):
+    """Class representing a tim (*.tim) file.
+
+    Attributes:
+        data: Dictionary with keys \"comment\" & time as float and value as List of floats".\n
+        serializer_config: TimSerializerConfig
+
+    Raises:
+        ValueError: If the time and values are not parsable to float, or when the amount of columns differs per timeseries.
+    """
```
priscavdsluis, line 21:
```suggestion
    """Class representing a tim (*.tim) file."""
```

The `Raises` does not need to be documented on class level, only on function level.
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html.

As a team we have decided to document the attributes of a class not on class level, but on attribute level.

So for example:

```python
    serializer_config = TimSerializerConfig()
    """TimSerializerConfig: The serialization configuration for the tim file."""

    comments: List[str]
    """List[str]: A list with the header comment of the tim file."""
    
    timeseries: Dict[float, List[float]]
    """Dict[float, List[float]]: A dictionary containing the time series. The keys are the times and the values are the values corresponding to that time."""

```

hydrolib/core/dflowfm/tim/models.py
```diff
@@ -0,0 +1,73 @@
...
+            if not firstlengthset:
+                length = len(v[time])
+                firstlengthset = True
+                continue
+
+            if length != len(v[time]):
+                raise ValueError(f"Problem with values in timeseries, for time {time}")
+            length = len(v[time])
+
+        return v
```
priscavdsluis, line 73:
```suggestion
        if len(v) == 0:
            return v
        
        values_iterator = iter(v.values())
        n_columns = len(next(values_iterator))
        
        if n_columns == 0:
            raise ValueError('Time series cannot be empty.')
        
        if not all(len(values) == n_columns for values in values_iterator):
            raise ValueError('All time series rows should have the same number of columns.')
    
        return v
```
or (for more detailed error messages):
```suggestion
        if len(v) == 0:
            return v
    
        timeseries_iterator = iter(v.items())
        _, columns = next(timeseries_iterator)
        n_columns = len(columns)
        
        if n_columns == 0:
            raise ValueError('Time series cannot be empty.')
        
        for time, values in timeseries_iterator:
            if len(values) != n_columns:
                raise ValueError(f'Time {time}: Expected {n_columns} columns, but was {len(values)}')
        
        return v
```

Please pay attention to clear error messages: informing the user that there is a problem, but not why, is not informative.

hydrolib/core/dflowfm/tim/models.py
```diff
@@ -0,0 +1,73 @@
...
+    def _timeseries_values(
+        cls, v: Dict[float, List[float]]
+    ) -> Dict[float, List[float]]:
+        """Validates if the amount of columns per timeseries match.
+
+        Args:
+            v (Dict[float, List[float]]): Value to validate, the timeseries in this case.
+
+        Raises:
+            ValueError: If the time and values are not parsable to float, or when the amount of columns differs per timeseries.
```
priscavdsluis, line 57:
```suggestion
            ValueError: When the amount of columns differs per timeseries.
```

You do not need to specify the type validation, it is not the responsibility of this function.

veenstrajelmer:
While doing a quick test I noticed the TimModel is not quite in line with other similar hydrolib-core objects. Since the timfile is similar than the files for XYNModel/XYZModel etc, I would expect more similarities in the object formatting. Below is some example code that shows how the objects are built up. In line with this, I would expect `TimModel` to have a `points` property and `TimModel.points` to be a list of `Points` (or similar object). Not set in stone of course, but it is an expectation which can be discussed.

For completeness, the `Timeseries` object (from a bc-file) is also included, but since that object is way more complex it might not be relevant.

```
import hydrolib.core.dflowfm as hcdfm

#xyz model
file_xyz = r'p:\archivedprojects\11205258-006-kpp2020_rmm-g6\C_Work\08_RMM_FMmodel\geometry_j19_NL_6-v2\rmm_vzm_v1p1_initial_water_level_onlyriv.xyz'
data_xyz = hcdfm.XYZModel(file_xyz)
print('\nXYZModel.points is list of XYZPoint, which has x/y/z/comment keys (first 5 are printed):') #TODO: comment can be dropped?
print(type(data_xyz.points))
print(data_xyz.points[:5])

#load xyn model
file_xyn = r'p:\archivedprojects\11206813-006-kpp2021_rmm-2d\C_Work\31_RMM_FMmodel\geometry_j19_6-v2\output_locations\rmm_vzm-j19_6-v2b_3_measurement_obs.xyn'
data_xyn = hcdfm.XYNModel(file_xyn)
print('\nXYNModel.points is list of XYNPoint, which has x/y/n keys (first 5 are printed):')
print(type(data_xyn.points))
print(data_xyn.points[:5])

#load pol/tek/pli/ldb file
file_pli = r'p:\11208054-004-dcsm-fm\models\model_input\bnd_cond\pli\DCSM-FM_OB_all_20181108.pli'
polyfile_object = hcdfm.PolyFile(file_pli)
poly_object_first = polyfile_object.objects[0]
print('\nPolyObject.points is list of Point, which has x/y/z/data keys (first 5 are printed):')
print(type(poly_object_first.points))
print(poly_object_first.points[:5])

# #bcfile
# file_bc_ts = r'c:\DATA\dfm_tools_testdata\hydrolib_bc\DCSM\steric_DCSM-FM_OB_all_20181108_firstpart.bc'
# m = hcdfm.ForcingModel(file_bc_ts)
# timeseries_object = m.forcing[0]
# print('\nTimeseries.datablock is list of lists with [time, val] (first 5 are printed):')
# print(type(timeseries_object.datablock))
# print(timeseries_object.datablock[:5])

#timfile
file_tim = r'p:\archivedprojects\11205258-006-kpp2020_rmm-g6\C_Work\08_RMM_FMmodel\structures_toRTC\Algerakering_GateLowerEdgeLevel.tim'
data_tim = hcdfm.TimModel(file_tim)
print('\nTimModel.timeseries is dict with times as keys and rest of columns as list of values:')
print(type(data_tim.timeseries))
print(data_tim.timeseries)
```

hydrolib/core/dflowfm/tim/serializer.py
```diff
@@ -0,0 +1,120 @@
...
+    @staticmethod
+    def _serialize_comment_lines(data) -> List[str]:
+        commentlines = []
+        for comment in data["comments"]:
+            commentlines.append(f"#{comment}")
+        return commentlines
+
+    @staticmethod
+    def _serialize_timeseries_lines(data, config) -> List[str]:
+        format_float = lambda v: f"{v:{config.float_format}}"
```
priscavdsluis, line 56: Why not define this lambda in the `_serialize_to_timeseries_block` method?

priscavdsluis (changes requested):
Some small remarks

hydrolib/core/dflowfm/tim/serializer.py
MRVermeulenDeltares: since _serialize_to_timeseries_block only needs format_float, passing config in that method would be passing too much information into the method.

priscavdsluis: But then you could also just pass the `config.float_format` to the method and define the lambda inside the method. 
Passing functions normally is a bit of a code smell.
But since it is all internal/private, lets keep it like this.

priscavdsluis (approved):

Issue #365, opened 2022-10-10
support for str instead of Path


veenstrajelmer:
**Is your feature request related to a problem? Please describe.**
Currently, it is necessary to import Path whenever using hydrolib functions: ``from pathlib import Path``

**Describe the solution you'd like**
This is acceptable, but I can imagine it would be fairly easy to add support for plain strings also, by converting all strings to Path(str) in hydrolib. This saves a bit of a hassle. At least, if chosen not to add support, I would suggest to raise a more explanatory error than this:
``AttributeError: 'str' object has no attribute 'is_absolute'``

Instead, explicitly informing the user to wrap ``Path()`` around the provided string.

MRVermeulenDeltares:
in the FileModel class, the filepath variable can have a check on the input for path.
in the FileModel class, the save method can have a check on the input for path.
in the FileModel class, the constructor can have a check on the input for path.
Classes inheriting from the FileModel will have the path input resolved via the parent class.
e.g. the DIMR class inherits from the ParsableFileModel which in turn inherits from FileModel.

PR #469, opened 2023-03-16
Support for str besides Path

MRVermeulenDeltares:
Refs: #365

### DIFF
--- a/hydrolib/core/basemodel.py
+++ b/hydrolib/core/basemodel.py
@@ -1020,7 +1020,7 @@ class FileModel(BaseModel, ABC):
     def _change_to_path(filepath):
         if filepath is None:
             return filepath
-        if isinstance(filepath, Path):
+        elif isinstance(filepath, Path):
             return filepath
         else:
             return Path(filepath)
--- a/hydrolib/core/basemodel.py
+++ b/hydrolib/core/basemodel.py
@@ -1026,7 +1026,7 @@ class FileModel(BaseModel, ABC):
             return Path(filepath)
 
     @validator("filepath")
-    def _conform_filepath_to_hydrolib_standard(cls, value):
+    def _conform_filepath_to_pathlib(cls, value):
         return FileModel._change_to_path(value)
 
 
--- a/hydrolib/core/basemodel.py
+++ b/hydrolib/core/basemodel.py
@@ -22,6 +22,7 @@ from typing import (
     Tuple,
     Type,
     TypeVar,
+    Union,
 )
 from weakref import WeakValueDictionary
 
@@ -673,16 +674,17 @@ class FileModel(BaseModel, ABC):
     # Absolute anchor is used to resolve the save location when the filepath is relative.
     _absolute_anchor_path: Path = PrivateAttr(default_factory=Path.cwd)
 
-    def __new__(cls, filepath: Optional[Path] = None, *args, **kwargs):
+    def __new__(cls, filepath: Optional[Union[Path,str]] = None, *args, **kwargs):
         """Create a new model.
         If the file at the provided file path was already parsed, this instance is returned.
 
         Args:
-            filepath (Optional[Path], optional): The file path to the file. Defaults to None.
+            filepath (Optional[Union[Path,str]], optional): The file path to the file. Defaults to None.
 
         Returns:
             FileModel: A file model.
         """
+        filepath = FileModel._change_to_path(filepath)
         with file_load_context() as context:
             if (file_model := context.retrieve_model(filepath)) is not None:
                 return file_model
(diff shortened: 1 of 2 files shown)


hydrolib/core/basemodel.py
```diff
@@ -704,14 +706,16 @@ def __init__(
...
-            filepath (Optional[Path], optional): The file path. Defaults to None.
+            filepath (Optional[Union[Path,str]], optional): The file path. Defaults to None.
             resolve_casing (bool, optional): Whether or not to resolve the file name references so that they match the case with what is on disk. Defaults to False.
             recurse (bool, optional): Whether or not to recursively load the model. Defaults to True.
         """
         if not filepath:
             super().__init__(*args, **kwargs)
             return
 
+        filepath = FileModel._change_to_path(filepath)
```
priscavdsluis, line 717: I was looking at your PR because I was wondering how you solved it in the end. Because I had to use similar logic for the newly introduced path style parameter in PR #361.
I was wondering what are your thought of the following: https://github.com/Deltares/HYDROLIB-core/blob/0986fe1a92309661507828c89739d2eeb7aeb695/hydrolib/core/basemodel.py#L783.

Maybe we can add the file path validation to the UserInputValidation class. 

Moreover, I noticed that in the `_change_to_path` function below the `filepath` is checked for None, but the check above the function call (line 713) already handles this situation.

hydrolib/core/basemodel.py
```diff
@@ -673,16 +674,17 @@ class FileModel(BaseModel, ABC):
     # Absolute anchor is used to resolve the save location when the filepath is relative.
     _absolute_anchor_path: Path = PrivateAttr(default_factory=Path.cwd)
 
-    def __new__(cls, filepath: Optional[Path] = None, *args, **kwargs):
+    def __new__(cls, filepath: Optional[Union[Path, str]] = None, *args, **kwargs):
```
arthurvd, line 677: Consider a
```python
PathOrStr = Union[Path, str]
```
right above this class and use it in the other places

hydrolib/core/basemodel.py
```diff
@@ -1012,6 +1016,19 @@ def _load(self, filepath: Path) -> Dict:
     def __str__(self) -> str:
         return str(self.filepath if self.filepath else "")
 
+    @staticmethod
+    def _change_to_path(filepath):
+        if filepath is None:
+            return filepath
+        if isinstance(filepath, Path):
```
arthurvd, line 1023: here an `elif`?

tests/test_basemodel.py
arthurvd, line 1: I would either test this on `BaseModel` itself (so not `FMModel`), or place the entire test code in `tests/dflowfm/test_mdu.py`.

arthurvd (changes requested):
Nice implementation, exactly how I understood it yesterday. Please considers some minor changes (see comments).

hydrolib/core/basemodel.py
MRVermeulenDeltares: What do you mean exactly?

hydrolib/core/basemodel.py
MRVermeulenDeltares: done

hydrolib/core/basemodel.py
MRVermeulenDeltares: discussed

tests/test_basemodel.py
MRVermeulenDeltares: The changes cannot be tested on BaseModel, since they are implemented on FileModel, which inherits from basemodel.
The tests cannot make use of the FileModel class, since the constructor calls abstract classes from FileModel.
Thus FMModel was taken for the test.

I am unsure why this should be moved to the test_mdu.py, since the software changes have taken plance in the FileModel which is located in the basemodel.py

hydrolib/core/basemodel.py
MRVermeulenDeltares: done

hydrolib/core/basemodel.py
arthurvd: @MRVermeulenDeltares : can you briefly summarize what you and Prisca decided?

tests/test_basemodel.py
arthurvd: Discussed, changes not needed: str/path changes are in `FileModel` inside basemodel.py, so it's good to also test it in test_basemodel.py (even though the `FMModel` class is used as the test object)

hydrolib/core/basemodel.py
MRVermeulenDeltares: Discussed what was implemented for #365 and why the none checks were added to the _change_to_path method.
No further impact on issue #365 related to the UserInputValidation.
Discussion can be resolved.

PR #470, opened 2023-03-17
Include old and new obscrossection into MDU output

### LINKED ISSUE #461: Include ObservationCrossSectionModel in MDU's Output model
Currently, in `FMModel.output.obs.crsfile`, we store a `DiskOnlyFileModel`.
But actually, the observation cross section file can already be stored in a dedicated `ObservationCrossSectionModel` object.

The type of `crsfile` should be changed to `Optional[List[Union[ObservationCrossSectionModel, PolyFile]]`.

It is important that when `load`ing an existing file (possibly with both old and new obscrs files), the parser/constructor correctly detects the correct class/type for each file.

arthurvd:
Fixes #461


tests/dflowfm/test_mdu.py
```diff
@@ -84,10 +89,68 @@ def test_mdu_with_optional_sections(self):
...
+        assert len(fmmodel.output.crsfile[0].observationcrosssection) == 4
+        assert (
+            fmmodel.output.crsfile[0].observationcrosssection[0].name
+            == "ObservCross_Chg_T1"
+        )
+        assert (
+            len(fmmodel.output.crsfile[0].observationcrosssection[3].xcoordinates) == 2
+        )
+
+    def test_mixed_obs_crs_files(self):
```
tim-vd-aardweg, line 125: We have seen that this test gives a different result compared to actually reading a `.mdu` file where you list both the `test.pli` and `ObservationPoints_crs.ini`.

I think writing a test that does the above has several advantages:
- You test the package the way the user would use it.
- It is more resistant to refactorings. Assume we add an additional argument to the `Parser` in the source code. You would have to update this test as well. While if you simply read the mdu file, you do not have to change this test. In both cases (reading a `.mdu` and the test you suggest) test if we successfully create a `PolyFile` for the `.pli` and an `ObservationCrossSectionModel` from `ObservationPoints_crs.ini`
- It makes the test more readable (easier to follow what happens). Of course, this is subjective :)

tim-vd-aardweg: If it was your intention to test the `Parser` instead, I would have checked if the given input would return the correct `Document` instead.

tim-vd-aardweg (changes requested):
I would like to see a different (or additional) testcase :)

tests/dflowfm/test_mdu.py
arthurvd: I agree with you: I have removed this parser-based test and replaced it by three FMModel loading tests, similar to what you have done for the obs ini/xyn mixed testing. See newer commits.

tests/dflowfm/test_mdu.py
```diff
@@ -15,10 +15,19 @@
     ProcessFluxIntegration,
     VegetationModelNr,
 )
-from hydrolib.core.dflowfm.obscrosssection.models import ObservationCrossSectionModel
-from hydrolib.core.dflowfm.polyfile.models import PolyFile
+from hydrolib.core.dflowfm.obscrosssection.models import (
```
tim-vd-aardweg, line 18: Could you clean up the unused imports in this file? 👼

tim-vd-aardweg (changes requested):
Just some minor clean up, then we can move this issue to test!

tests/dflowfm/test_mdu.py
arthurvd: See new commit.

tim-vd-aardweg (approved):
Review OK!

Issue #459, opened 2023-03-10
Add support for 'old' obs.xyn files (observationpoints)


arthurvd:
**Is your feature request related to a problem? Please describe.**
Currently, obs (and also crs) files are not supported by hydrolib. Upon reading them, the resulting dataset is empty. As a 2D modeller, I was not aware that this filetype is considered old, since it is being widely used in our models and therefore not possible to deprecate yet. Furthermore, I was not aware there existed a new alternative for it.

**Describe the solution you'd like**
Add support for *_obs.xyn (*_crs.pli files is handled under separate issue #364)

tim-vd-aardweg:
Check if it can be subclassed from the xyz model.

tim-vd-aardweg:
https://content.oss.deltares.nl/delft3dfm1d2d/D-Flow_FM_User_Manual_1D2D.pdf Section F.2.2.2 describes the .xyn format.

veenstrajelmer:
Additional request: validate that there are no duplicate observation point names

arthurvd:
> ```
> 1.23  3.45  'some text with spaces' And everything beyond does not matter
> 1.23 3.45 Only_the_first_word and the test is ignored
> ```
> 
> Would be great if both of the above cases result in an error instead.
> 
> Additional request: validate that there are no duplicate observation point names

@tim-vd-aardweg : if you're also ok with extending the scope of this issue a bit, let's keep Jelmer's two requests separate (and add @priscavdsluis's request too):
1. the first one under this same issue: raise an error from the parser if there is any text remaining on a line after successfully parsing the x+y+name already.
2. the validation of station name uniqueness now has an issue of its own: #483.
3. Finally, @priscavdsluis correctly remarked that api ref .md files must be added for the newly added `xyn` package scope. OK to that under this same issue.

veenstrajelmer:
Thanks for the implementation. There are still some issues that I encounter
```
import hydrolib.core.dflowfm as hcdfm
from pathlib import Path
file_xyn = r'p:\archivedprojects\11206813-006-kpp2021_rmm-2d\C_Work\31_RMM_FMmodel\geometry_j19_6-v2\output_locations\rmm_vzm-j19_6-v2b_3_measurement_obs.xyn'
```

XYNModel exists, but is not exposed via init file, therefore the second line fails:
```
hcdfm.xyn.models.XYNModel(Path(file_xyn))
hcdfm.XYNModel(Path(file_xyn))
```

[...]

PR #472, opened 2023-03-17
feat: Add support for xyn files

tim-vd-aardweg:
Fixes: #459


tim-vd-aardweg:
There are 2 code smells left that I believe we can ignore. I do not like to put those strings in a constant.

hydrolib/core/dflowfm/xyn/name_extrator.py
```diff
@@ -0,0 +1,77 @@
+class NameExtractor:
```
arthurvd, line 1: Maybe this whole class could be replaced by the following regexp?
```python
prog=re.compile("^\s*('?)([^'#]+)\\1\s*#?.*$")
if (match := prog.match(text) is not None):
    return match.group(2)
# ...
```
See here some examples on which it works:
```
In [30]: print(prog.match("'somevalue in quotes' #comment").group(2))
somevalue in quotes

In [31]: print(prog.match("  somevalue_without_quotes #comment").group(2))
somevalue_without_quotes

In [32]: print(prog.match("  somevalue_without_quotes_and_no_comment  ").group(2))
somevalue_without_quotes_and_no_comment

In [33]: print(prog.match("'somevalue in quotes, but no comment'  ").group(2))
somevalue in quotes, but no comment

In [34]: print(prog.match("'somevalue in quotes, but no comment and missing closing quote  "))
None
```

arthurvd: Upon second checking: no comments at the end of a data line are supported at all in .xyn files, so please remove the `#` handling (either in your own code, or in my above suggestion).

hydrolib/core/dflowfm/xyn/serializer.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import Dict
+
+from hydrolib.core.basemodel import SerializerConfig
+
+
+class XYNSerializer:
+    @staticmethod
+    def serialize(path: Path, data: Dict, config: SerializerConfig) -> None:
+        """
+        Serializes the XYN data to the file at the specified path.
```
arthurvd, line 11:
```suggestion
        Serializes the observation point data to an .xyn file at the specified path.
```

tests/data/input/obsfile_cases/single_ini/obs.ini
arthurvd, line 1: this .ini duplicates with the case with both xyn and ini. Maybe only have the mdu's different, and refer to the same .ini?

tests/data/input/obsfile_cases/single_xyn/obs.xyn
arthurvd, line 1: this duplicates with the xyn from the case with both ini and xyn. Maybe only include the MDU and refer to the existing xyn file?

arthurvd (changes requested):
Hi @tim-vd-aardweg , please see my review comments.

tests/data/input/obsfile_cases/single_ini/obs.ini
tim-vd-aardweg: I am a bit conflicted here. I generally do not want to use the  same test data (file) for different tests (with different purposes). Let's discuss it today.

tim-vd-aardweg: Discussed this with Arthur and we will leave it as is for now.

tests/data/input/obsfile_cases/single_xyn/obs.xyn
tim-vd-aardweg: Discussed this with Arthur and we will leave it as is for now.

hydrolib/core/dflowfm/xyn/name_extrator.py
tim-vd-aardweg: I have spent some time trying to get it to work with a single regex, but I could not get it to work :(

hydrolib/core/dflowfm/xyn/models.py
```diff
@@ -46,7 +46,7 @@ def _ext(cls) -> str:
 
     @classmethod
     def _filename(cls) -> str:
-        return "sample"
+        return "stations_obs.xyn"
```
arthurvd, line 49:
```suggestion
        return "stations_obs"

    @abstractclassmethod
    def _ext(cls) -> str:
        return ".xyn"
```

arthurvd (changes requested):
All fine now, except for one small new thing that I overlooked last time. Can you consider/commit my suggestion? After that I will merge.

arthurvd (approved):

Issue #446, opened 2023-01-30
Implement meteo for external forcing


xldeltares:
**Is your feature request related to a problem? Please describe.**
In project Santa Cruz, we need to implement rainfall on 2D grid as meteo forcing. For now hydrolib-core does not have the meteo implemented. 

**Describe the solution you'd like**
a new class Meteo(INIBasedModel) that writes a `[Meteo]` block for use inside an external forcings file.

xldeltares:
@arthurvd: That would be great if you still could review/finish it in this sprint! I will create a PR and please feel free to work on it. Thanks!

arthurvd:
@priscavdsluis / @tim-vd-aardweg / @MRVermeulenDeltares : our discussion about the diskOnlyFileModel will be continued under a new issue: #478.

priscavdsluis:
I reviewed the PR and the changes look good. 
However there is the property `forcing` on the new Meteo object. I want to verify first if it is correct and find an example with non-global meteo data.

PR #477, opened 2023-03-23
#446: Implement meteo for external forcing

xldeltares:


hydrolib/core/dflowfm/ext/models.py
```diff
@@ -181,6 +184,117 @@ def validate_location_type(cls, v: str) -> str:
...
+    attribute in Meteo class.
+    """
+
+    bcascii = "bcAscii"
+    """str: Space-uniform time series in <*.bc> file."""
+
+    netcdf = "netcdf"
+    """str: NetCDF, either with gridded data, or multiple station time series."""
+
+    uniform = "uniform"
```
priscavdsluis, line 199: To check: I do not see this option in the manual

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -181,6 +184,117 @@ def validate_location_type(cls, v: str) -> str:
...
+    _disk_only_file_model_should_not_be_none = (
+        validator_set_default_disk_only_file_model_when_none()
+    )
+
+    _header: Literal["Meteo"] = "Meteo"
+    quantity: str = Field(alias="quantity")
+    forcingfile: ForcingModel = Field(alias="forcingFile")
+    forcingfiletype: MeteoForcingFileType = Field(alias="forcingFileType")
+    targetmaskfile: Optional[PolyFile] = Field(None, alias="targetMaskFile")
+    targetmaskinvert: Optional[bool] = Field(None, alias="targetMaskInvert")
```
priscavdsluis, line 260: This field should be written as `yes` or `no`, according to the manual

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -181,6 +184,117 @@ def validate_location_type(cls, v: str) -> str:
...
+
+    forcingfiletype_validator = get_enum_validator(
+        "forcingfiletype", enum=MeteoForcingFileType
+    )
+    interpolationmethod_validator = get_enum_validator(
+        "interpolationmethod", enum=MeteoInterpolationMethod
+    )
+
+    @property
+    def forcing(self) -> Optional[ForcingBase]:
```
priscavdsluis, line 276: We could remove this property as I assume it only works with a `ForcingModel`.
We can add it once it is a requested functionality and think about a way to do this.

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -181,6 +184,117 @@ def validate_location_type(cls, v: str) -> str:
...
+
+    comments: Comments = Comments()
+
+    _disk_only_file_model_should_not_be_none = (
+        validator_set_default_disk_only_file_model_when_none()
+    )
+
+    _header: Literal["Meteo"] = "Meteo"
+    quantity: str = Field(alias="quantity")
+    forcingfile: ForcingModel = Field(alias="forcingFile")
```
priscavdsluis, line 257: Given there are 3 supported forcing file types, this should be a Union of this several file types.

tests/dflowfm/test_ext.py
```diff
@@ -584,6 +590,53 @@ def test_given_dict_values_doesnot_raise(self, dict_values: dict):
...
+            with pytest.raises(ValidationError) as error:
+                Meteo(**dict_values)
+
+            expected_message = f"{alias_field}\n  field required "
+            assert expected_message in str(error.value)
+
+        def test_construct_from_file(self):
+            input_ext = (
+                test_input_dir
+                / "e02/f006_external_forcing/c069_rain_bc/rainschematic.ext"
```
priscavdsluis, line 629: The additional test files (like the dimr.xml) could be removed from the repository. But I will leave it for now since we are going to change our testing approach with regards to test data anyway.

hydrolib/core/dflowfm/ext/models.py
arthurvd: I have fixed this now in the User Manual.

hydrolib/core/dflowfm/ext/models.py
arthurvd: You are right about the manual. But kernel supports also 0/1 (and much more) for this field.
And most importantly: for many of our other 0/1 settings present in hydrolib-core it *must* be a 0/1 value for the kernel, so let's keep that here as well.

By the way: this particular field is not yet in the GUI plugin: it gets imported, probably dropped, and not written upon a new export. So also there 0/1 is fine.

tests/dflowfm/test_ext.py
arthurvd: I agree!

hydrolib/core/dflowfm/ext/models.py
arthurvd: ok, done in newer commit.

hydrolib/core/dflowfm/ext/models.py
priscavdsluis: Ok good to know, would it be handy to update the manual? 
Most values are written as 0/1 so no/yes seems out of the ordinary

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -254,7 +255,9 @@ class Comments(INIBasedModel.Comments):
 
     _header: Literal["Meteo"] = "Meteo"
     quantity: str = Field(alias="quantity")
-    forcingfile: ForcingModel = Field(alias="forcingFile")
+    forcingfile: Union[ForcingModel, TimModel, DiskOnlyFileModel] = Field(
```
arthurvd, line 258: At first sight, this looks good, but unfortunately a unit test with a `.tim` file fails. See my recent commit.

arthurvd (changes requested):
It does not yet work correctly for a `.tim`. See my recent commit 350bf28, for an example model with failing unit test.

hydrolib/core/dflowfm/ext/models.py
priscavdsluis: The provided tim file was incorrect, so pydantic decides that it is probably not a `TimModel`, but the next one that does not fail: `ForcingModel`. The tim file cannot contain duplicate times and text after the columns.
I also switched the order of the union, as `ForcingModel` seems to be less strict.

arthurvd: ahh, apologies for causing confusion!

arthurvd (approved):

PR #490, opened 2023-03-29
Add 4 missing 1D2D settings to MDU (#489)

### LINKED ISSUE #489: Some 1D2D settings not yet supported in MDU file
**Describe the bug**
When I tried to convert an existing MDU file into an FMModel class multiple settings are not read. This has led to disappearing settings after using the ".copy()" functionality and writing the FMModel back to an .mdu file. Some of the settings which disappear are:
```
[numerics]
EpsMaxlev = 1.23e-4 # Stop criterium for non linear iteration
EpsMaxlevm = 1.23e-4 # Stop criterium for Nested Newton loop in non linear iteration

[time]
AutoTimestepNoStruct = 0 # or 0 = no, 1 = yes (Exclude structure links (and neighbours) from time step limitation). Default: 0.
AutoTimestepNoQout = 1 # 0 = no, 1 = yes (Exclude negative qin terms from time step limitation). Default: 1.
```
The disappearance of these settings has caused a previously running model to crash after it has been read and written by hydrolib-core.

**To Reproduce**
1. load an `FMModel` with and MDU file containing above keywords
2. Copy the model to a second one.
3. Save the second one
4. Notice disappeared settings in saved file.

**Expected behavior**
The above settings should be kept, also upon re-saving.

I ran my scripts using D-Hydro2022.04 1D2D and Hydrolib-core==0.3.1 (but the same issue seems to occur in the latest 0.4.1)

**Additional context**
Reported by Ivar Lokhorst MSc, Royal HaskoningDHV

arthurvd:
Fixes #489.


hydrolib/core/dflowfm/mdu/models.py
```diff
@@ -561,6 +570,14 @@ class Comments(INIBasedModel.Comments):
         dtinit: Optional[str] = Field(
             "Initial timestep in seconds [s].", alias="dtInit"
         )
+        autotimestepnostruct: Optional[str] = Field(
+            "Exclude structure links (and neighbours) from time step limitation (0 = no, 1 = yes).",
```
MRVermeulenDeltares, line 574: These two comments do not seem in line with the comments stated in #489, are they correct?

[time]
AutoTimestepNoStruct = 0 **# or 0 = no, 1 = yes (Exclude negative qin terms from time step limitation). Default: 0**.
AutoTimestepNoQout = 1 **# 0 = no, 1 = yes (Exclude structure links (and neighbours) from time step limitation). Default: 1.**

tests/data/reference/fm/special_3d_settings.mdu
```diff
@@ -157,6 +159,8 @@
     dtNodal                 = 21600.0  # Time interval (s) for updating nodal factors in astronomical boundary conditions
     dtMax                   = 120.0    # Maximal computation timestep (s)
     dtInit                  = 60.0     # Initial computation timestep (s)
+    AutoTimestepNoStruct    = 0        # Exclude structure links (and neighbours) from time step limitation (0 = no, 1 = yes).
```
MRVermeulenDeltares, line 162: If the above is changed, this will also need to be changed.

MRVermeulenDeltares (changes requested):

hydrolib/core/dflowfm/mdu/models.py
MRVermeulenDeltares: Incorrectly state in https://github.com/Deltares/HYDROLIB-core/issues/489, at that time, adjusted in lateron, implementation correct

tests/data/reference/fm/special_3d_settings.mdu
MRVermeulenDeltares: see above

MRVermeulenDeltares (approved):

PR #504, opened 2023-04-03
450: Update MeshKernel to latest v2.0.2

priscavdsluis:
Note that this MeshKernel version does not have MacOS support, so HYDROLIB-core will discontinue MacOS support for now.


priscavdsluis:
Removed the MacOS from the workflows.

arthurvd (approved):
Looks good, I verified that there are no other GitHub actions that need changes and there were also no other meshkernel API changes apart from the create that Prisca already fixed. Ready for merging.

veenstrajelmer:
This change is tested and it successfully installs both hydrolib-core and meshkernel 2.0.2

Issue #497, opened 2023-03-30
Reading StructureModel with timfiles crashes


veenstrajelmer:
**Is your feature request related to a problem? Please describe.**
When testing with a `structures.ini` file that contains timefiles with for instance GateLowerEdgeLevels, there is a long list of validation errors about many unrelated fields (friction, length, shift, csDefId, bridge). When using a comparable file with fixed values instead of the timfiles, the file is read in successfully. I did not expect yet that the timeseries from the timfile was correctly parsed, since https://github.com/Deltares/HYDROLIB-core/issues/348 is still open. However, there seems to be more that is going wrong here.

**To Reproduce**
```
import hydrolib.core.dflowfm as hcdfm
#file_struct = r'p:\archivedprojects\11205258-006-kpp2020_rmm-g6\C_Work\08_RMM_FMmodel\structures_toRTC\RMM_structures_open_cl10_coeff10.ini' #without timfiles
file_struct = r'p:\archivedprojects\11205258-006-kpp2020_rmm-g6\C_Work\08_RMM_FMmodel\structures_toRTC\RMM_structures_ts_cl10_coeff10.ini' #with timfiles
structs = hcdfm.StructureModel(file_struct)
```

Gives this validation error (only for structure 21, 
```
RMM_structures_ts_cl10_coeff10.ini -> structure -> 21 -> HY_17.96_R_SK_Kering_Hollandsche_IJssel -> type
  unexpected value; permitted: 'weir' (type=value_error.const; given=generalStructure; permitted=('weir',))
RMM_structures_ts_cl10_coeff10.ini -> structure -> 21 -> HY_17.96_R_SK_Kering_Hollandsche_IJssel -> __root__
  Specify location either by setting `branchId` and `chainage` or `num/x/yCoordinates` or `polylinefile` fields. (type=assertion_error)
RMM_structures_ts_cl10_coeff10.ini -> structure -> 21 -> HY_17.96_R_SK_Kering_Hollandsche_IJssel -> type
  unexpected value; permitted: 'universalWeir' (type=value_error.const; given=generalStructure; permitted=('universalWeir',))
RMM_structures_ts_cl10_coeff10.ini -> structure -> 21 -> HY_17.96_R_SK_Kering_Hollandsche_IJssel -> allowedFlowDir
  field required (type=value_error.missing)
[...]

PR #517, opened 2023-04-07
#497: Add support for tim files in the structure file

priscavdsluis:


tests/utils.py
```diff
@@ -146,3 +148,64 @@ def error_occurs_only_once(error_message: str, full_error: str) -> bool:
...
+
+    Args:
+        content (str): The content of the file as string.
+        filename (str): The file name.
+
+    Example:
+        >>>     with create_temp_file("some_file_content", "some_file_name") as temp_file:
+        >>>         print(f"Do something with {temp_file}")
+
+    Returns:
```
arthurvd, line 165:
```suggestion
    Yields:
```

tests/utils.py
```diff
@@ -146,3 +148,64 @@ def error_occurs_only_once(error_message: str, full_error: str) -> bool:
...
+
+    Args:
+        content (str): The content of the file as list of string (lines of the file).
+        filename (str): The file name.
+
+    Example:
+        >>>     with create_temp_file_from_lines(["some_file_content"], "some_file_name") as temp_file:
+        >>>         print(f"Do something with {temp_file}")
+
+    Returns:
```
arthurvd, line 188:
```suggestion
    Yields:
```

tests/utils.py
```diff
@@ -146,3 +148,64 @@ def error_occurs_only_once(error_message: str, full_error: str) -> bool:
         return False
 
     return full_error.count(error_message) == 1
+
+
+@contextmanager
+def create_temp_file(content: str, filename: str) -> Generator[Path, None, None]:
```
arthurvd, line 154: Mooi!

tests/utils.py
```diff
@@ -146,3 +148,64 @@ def error_occurs_only_once(error_message: str, full_error: str) -> bool:
...
+    Returns:
+        Generator[Path, None, None]: Generator with the path to the file in the temporary directory as yield type.
+    """
+    content = "\n".join(lines)
+    with create_temp_file(content, filename) as file:
+        yield file
+
+
+@contextmanager
+def get_temp_file(filename: str) -> Generator[Path, None, None]:
```
arthurvd, line 197: Can you not do without this function, and on the (single) call site in `create_temp_file(...)`, use `tempfile.NamedTemporaryFile(...)` instead?

tests/dflowfm/test_structure.py
```diff
@@ -1908,6 +1920,34 @@ def _create_pump_values(self) -> dict:
 
 
 class TestGeneralStructure:
+    def test_initialize_gateopeningwidth_with_timfile_initializes_timmodel(self):
+        gateloweredgelevel = test_input_dir / "tim" / "triple_data_for_timeseries.tim"
```
arthurvd, line 1924:
```suggestion
        gateloweredgelevel = test_input_dir / "tim" / "single_data_for_timeseries.tim"
```
Power of this test does not change by the above, but the input makes more sense with a single timeseries for a single structure parameter.

arthurvd (changes requested):
Code changes look good, only see some minor suggestions I made.

One remaining question: why does the Union type work correctly here, whereas it does not yet work well in the meteo issue with timeseries (See my comment there on: https://github.com/Deltares/HYDROLIB-core/pull/477#discussion_r1160667306 )
Do you know why?

priscavdsluis:
> Code changes look good, only see some minor suggestions I made.
> 
> One remaining question: why does the Union type work correctly here, whereas it does not yet work well in the meteo issue with timeseries (See my comment there on: [#477 (comment)](https://github.com/Deltares/HYDROLIB-core/pull/477#discussion_r1160667306) ) Do you know why?

@arthurvd, I just checked the issue with the meteo block.
The provided tim file is incorrect, so pydantic decides that it is probably not a `TimModel`, but the next one that does not fail: `ForcingModel`

tests/dflowfm/test_structure.py
priscavdsluis: Done

tests/utils.py
priscavdsluis: All credits to @tim-vd-aardweg :sunglasses:

tests/utils.py
priscavdsluis: Done

tests/utils.py
priscavdsluis: Done

tests/utils.py
arthurvd: ok to postpone, but I don't think a wrapper class would be needed. It would be a one-liner change on the call site, if I'm seeing this correctly.

arthurvd (approved):

Issue #529, opened 2023-04-13
Some symbols in comments are not parsed correctly, and as a result not saved correct.


priscavdsluis:
**Describe the bug**
Special symbols such as `ζ`, `°` or `Δ` are not parsed correctly. This results in an incorrect file when saved again.

**To Reproduce**
```python
        from hydrolib.core.dflowfm import FMModel
        path = "fm.mdu"

        model_a = FMModel()
        model_a.filepath = path
        model_a.save()

        model_b = FMModel(filepath=path)
        model_b.save()
```
The model is first saved correctly, but after loading and saving it again the special symbols look weird:
![image](https://user-images.githubusercontent.com/36264671/231749388-ae2c1baf-6d82-4338-9577-8393a725743d.png)

**Failing test to reproduce**
```python
    def test_symbols():
        from hydrolib.core.dflowfm import FMModel
        path = "fm.mdu"

        model_a = FMModel()
        assert model_a.physics.comments.initialtemperature == "Initial temperature [◦C]."
        model_a.filepath = path
        model_a.save()

        model_b = FMModel(filepath=path)
        assert model_b.physics.comments.initialtemperature == "Initial temperature [◦C]."
        model_b.save()
```

**Expected behavior**
The comments when first saving and when saving after loading should remain the same.

**Version info (please complete the following information):**
 - Version 0.4.1

PR #534, opened 2023-04-14
#529: Special characters should be parsed correctly from file

priscavdsluis:
Fixes: #529 

Also made sure that a saved and loaded default FMModel are equal. 
Correctly reading the special characters were the first part, and the second part was the reading/writing of empty comments.


tim-vd-aardweg (commented):
Overall, the changes look good!

Should we also:
- Update the `write_ini` function in `hydrolib\core\dflowfm\ini\serializer.py` to:

```python
with path.open("w", encoding="utf8") as f:
    for line in serializer.serialize(document):
        f.write((line + "\n"))
```

- Update the `serialize` function in `hydrolib\core\dimr\serializer.py` to:

```python
with path.open("w", encoding="utf8") as f:
    f.write(xml)
```

- Update the `serialize` function (line 12) in `hydrolib\core\rr\topology\serializer.py` to:
```python
with path.open("w", encoding="utf8") as f:
    for node in data["node"]:
	line = (
		f"NODE {NodeFileSerializer._to_line(node, config)} node{os.linesep}"
		)
	 f.write(line)
```

- Update the `serialize` function (line 57) in `hydrolib\core\rr\topology\serializer.py` to:
```python
with path.open("w", encoding="utf8") as f:
    for link in data["link"]:
        line = f"BRCH {LinkFileSerializer._to_line(link)} brch{os.linesep}"
	f.write(line)
```

- Update the `assert_files_equal` function in `tests\utils.py` to:
```python
with file.open(encoding="utf-8") as af:
    actual_lines = af.readlines()

with reference_file.open(encoding="utf-8") as rf:
    reference_lines = rf.readlines()
```

- Update the `create_temp_file` function in `tests\utils.py` to:
```python
with get_temp_file(filename) as file:
    with open(file, "w", encoding="utf-8") as f:
 	f.write(content)
    yield file
```

priscavdsluis:
@tim-vd-aardweg I updated the test utils and also some of the parser/serializer, except the DIMR serializer. 
This was a bit tricky, because in the xml file the encoding is also written because of the following:

```python
xml = to_string.toprettyxml(indent="  ", encoding="utf-8")
```

Changing this will cause a failing test.

tim-vd-aardweg (approved):
Review OK!

Issue #535, opened 2023-04-14
Update oldextfile header


veenstrajelmer:
**What is the need for this task.**
Extfile header in dflowfm git and hydrolib-core are not aligned and/or up to date.

**What is the task?**
Extfile header is copied from https://git.deltares.nl/oss/delft3d/-/blob/main/src/engines_gpl/dflowfm/res/example.ext to hydrolib-core so it is also added to the oldextfile. However, the ordering was changed a bit to be in line with the FM kernel by @arthurvd. Later, it was discovered that the keyword `nudge_salinity_temperature was not in this file, so this still has to be added. Would be good to also update the `example.ext` on the dflowfm git.

priscavdsluis:
Quantity `nudge_salinity_temperature` was added in PR #536

MAfarrag:
@priscavdsluis @veenstrajelmer can you please check if this issue is still relevant or not ?

veenstrajelmer:
I guess since the old extfile will be phased out soon, this issue can be closed as won't do. Do you agree?

PR #536, opened 2023-04-14
#535: Add the quantity `nudge_salinity_temperature` to ext old file header

priscavdsluis:
Related to issue #535.
Partially resolves the issue.


tim-vd-aardweg (approved):
Review OK!

PR #580, opened 2023-11-02
570 add computedairdensity mdu keywords

veenstrajelmer:


hydrolib/core/dflowfm/mdu/models.py
```diff
@@ -564,6 +567,7 @@ class Comments(INIBasedModel.Comments):
     windpartialdry: bool = Field(True, alias="windPartialDry")
     pavbnd: float = Field(0.0, alias="pavBnd")
     pavini: float = Field(0.0, alias="pavIni")
+    computedairdensity: int = Field(0, alias="computedAirdensity")
```
priscavdsluis, line 570: Why is this an `int` instead of a `bool` (like the other changes)?

priscavdsluis (approved):

PR #589, opened 2023-11-20
Use Pydantic 2, but use V1 API.

### LINKED ISSUE #588: Migrate to Pydantic V2
Over at https://github.com/Deltares/Ribasim/pull/731 we migrated to Pydantic v2. Users combining hydrolib(-core) with ribasim (via HydroMT) will now inevitably hit environment problems, as we're using different pydantic versions.

evetion:
Fixes #588 

Fixes https://github.com/Deltares/Ribasim/issues/822


evetion:
@arthurvd @priscavdsluis Can we review/merge/release this? It would be good if we can use ribasim & hydrolib(-core) together again.

priscavdsluis (approved):

PR #601, opened 2024-01-31
add py311 to ci.yml and pyproject.toml

veenstrajelmer:


README.md
```diff
@@ -2,7 +2,7 @@
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
 [![ci](https://github.com/Deltares/HYDROLIB-core/actions/workflows/ci.yml/badge.svg)](https://github.com/Deltares/HYDROLIB-core/actions/workflows/ci.yml)
 [![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=Deltares_HYDROLIB-core&metric=alert_status)](https://sonarcloud.io/dashboard?id=Deltares_HYDROLIB-core)
-
+[![Supported versions](https://img.shields.io/pypi/pyversions/hydrolib-core.svg)](https://pypi.org/project/hydrolib-core)
```
veenstrajelmer, line 5: I propose to add this badge to the readme.md: https://img.shields.io/pypi/pyversions/hydrolib-core.svg

priscavdsluis:
Replaced our `str, Enum` mixins with the `StrEnum` from the `strenum` package.
The built-in module `enum` also offers an `StrEnum`, but this class is only available from python >= 3.11.
Since we also support earlier versions, I decided to use the `strenum` package, which offers the same functionality across all python versions, without breaking changes such as using the mixin. 

https://blog.pecar.me/python-enum

hydrolib/core/dflowfm/extold/models.py
```diff
@@ -3,6 +3,7 @@
 from typing import Any, Callable, Dict, List, Optional, Union
```
MRVermeulenDeltares, line 3: is the from enum import enum still used above^?

hydrolib/core/dflowfm/structure/models.py
```diff
@@ -11,6 +11,7 @@
```
MRVermeulenDeltares, line 11: This class appears to still make use of from enum import Enum.
mainly because of the DambreakAlgorithm(int, Enum).
Should this also be updated to IntEnum as similar to the StrEnum, or does this not cause any issues?

priscavdsluis: Only the str, Enum mixins were giving problems, so we should only update these.

MRVermeulenDeltares (approved):
I think the change is good, since the StrEnum package can be used in version prior to 3.11.
Git actions also seems to run fine on version 3.8 till 3.11.
Please have a look at my last comment, once that has been resolved the PR is Ok in my opinion.

tim-vd-aardweg (approved):

pyproject.toml
```diff
@@ -20,6 +20,7 @@ numpy = "^1.21"
 pydantic = "^2.5"
 lxml = "^4.6"
 meshkernel = "^4.0.2"
+strenum = "^0.4.15"
```
veenstrajelmer, line 23: This is the latest version, please consider loosening this constraint to avoid dependency conflicts when combining it with other packages in an environment.

hydrolib/core/dflowfm/extold/models.py
priscavdsluis: Removed it

pyproject.toml
priscavdsluis: Updated version requirement for strenum to ^0: >= 0.0.0, <1.0.0

PR #613, opened 2024-02-16
611 not possible to refine meshes with cell sizes smaller than 10 m

veenstrajelmer:


RuudHurkmans (approved):
I agree with the fix! From our point of view, 10 m is already rather course, so we have to be aware of the default and always set a values.

Issue #562, opened 2023-09-15
dimr_config for parallel run is incorrect


veenstrajelmer:
**Describe the bug**
When generating a dimr_config.xml with hydrolib for an MPI run, we need to pass the `process` and `mpiCommunicator` arguments. The expected dtype for `process` is `int`, but this should not be directly passed but converted to a space-separated range. For 4 cores, the entry in the dimr should be `<process>0 1 2 3</process>`, but instead it reads `<process>4</process>`.

**To Reproduce**
The file resulting from the code below contains an invalid `<process>` entry, which makes the model run crash:
```
from hydrolib.core.dimr.models import DIMR, FMComponent
fm_comp = FMComponent(name="test", workingDir='.', inputfile='test.mdu',
                      process=4, mpiCommunicator="DFM_COMM_DFMWORLD")
dimr = DIMR(component=fm_comp)
dimr.save(filepath='dimr_config.xml')
```

**Expected behavior**
`<process>0 1 2 3</process>` in the dimr_config.xml. It might be necessary to align processor/core numbers between different components of the DIMR file, but that is currently unsure. This must also be fixed for reading the xml file

**Version info (please complete the following information):**
 - OS: Windows
 - Version 0.5.2

**Additional request**
Since the `mpiCommunicator` is always the same ("DFM_COMM_DFMWORLD"), we could also consider to always write it to the dimr_config if the amount of processes is >1. One argument less, always nice.

tim-vd-aardweg:
So, just to be sure. If the processes = 5, we should write: `<process>0 1 2 3 4</process>`?

Should we still support the files that we have written incorrectly? In other words, should we still be able to parse: `<process>4</process>`?

@veenstrajelmer & @rhutten is this correct?

MRVermeulenDeltares:
Currently I have assumed the following:

This change is only for FM and not RR.

When 0 is input for process the dimr_config.xml will not have <process> added.

When 1 is input for process the dimr_config.xml will have <process>0</process> added.

open questions:
What should happen if an user inputs something different than a int?
e.g. 
```
    component = FMComponent(
        name="test",
        workingDir=".",
        inputfile="test.mdu",
        process="Four",
        mpiCommunicator="DFM_COMM_DFMWORLD",
    )
```
	
or

```
    component = FMComponent(
        name="test",
        workingDir=".",
        inputfile="test.mdu",
        process=4.0,
        mpiCommunicator="DFM_COMM_DFMWORLD",
    )
[...]

PR #623, opened 2024-04-17
Bug: dimr_config for parallel run is incorrect

MRVermeulenDeltares:


tim-vd-aardweg:
Fixes: #562

tim-vd-aardweg:
Have a look at the SonarCloud issue please! :)

hydrolib/core/dimr/models.py
```diff
@@ -52,7 +52,7 @@ class Component(BaseModel, ABC):
     name: str
     workingDir: Path
     inputFile: Path
-    process: Optional[int]
```
tim-vd-aardweg, line 55: This should remain an `int`, right? We only want to ensure that, for example, the integer 3 is written as `<process>0 1 2</process>` and not as `3`?

We still want to be able to do:
```
fm_comp = FMComponent(**data)
fm_comp.process = 4
```

Or did I missunderstand the issue?

hydrolib/core/dimr/models.py
```diff
@@ -88,6 +88,16 @@ class FMComponent(Component):
 
     library: Literal["dflowfm"] = "dflowfm"
 
+    @validator("process", pre=True, allow_reuse=True)
```
tim-vd-aardweg, line 91: Is the allow_reuse really needed?

tests/test_model.py
```diff
@@ -522,3 +522,117 @@ def test_model_diskonlyfilemodel_field_is_constructed_correctly(
...
+    ],
+)
+def test_dimr_with_fmcomponent_saving_process(
+    tmp_path, input_process: int, expected_process_format: str
+):
+    component = FMComponent(
+        name="test",
+        workingDir=".",
+        inputfile="test.mdu",
+        process=input_process,
```
tim-vd-aardweg, line 544: Currently `process` is (incorrectly??) typehinted as an `Optional[str]`, but here you pass an `int`

tests/test_model.py
```diff
@@ -522,3 +522,117 @@ def test_model_diskonlyfilemodel_field_is_constructed_correctly(
...
+@pytest.mark.parametrize(
+    "input_process",
+    [
+        pytest.param("lalala"),
+        pytest.param("123"),
+        pytest.param(0.5),
+    ],
+)
+def test_fmcomponent_process_after_init_with_incorrect_input_throws_valueerror(
+    input_process: int,
```
tim-vd-aardweg, line 624: Last thing you _could_ check if you really wanted, is the same as this test, but instead of setting the `process` in the constructor, you set the attribute after instantiation:

`component.process = input_process`

MRVermeulenDeltares: will update the tests

tests/test_model.py
MRVermeulenDeltares: 0 1 2 3  on itself cannot be an int right, which means it should be a string. or a list of ints?

MRVermeulenDeltares: _set_process_correctly will return the string

hydrolib/core/dimr/models.py
MRVermeulenDeltares: Not needed, by default true will remove

hydrolib/core/dimr/models.py
MRVermeulenDeltares: yes, you should be able to set it like this when using the model, but when it is read from the file it should be the correct format and be "0 1 2 3" when 4 is the input.
I think this should be an union of int and str in that case?
Since the input can be either int or str and the output is always str, since "0 1 2 3" cannot be of type int.

hydrolib/core/dimr/models.py
```diff
@@ -88,6 +88,78 @@ class FMComponent(Component):
 
     library: Literal["dflowfm"] = "dflowfm"
 
+    @validator("process", pre=True)
+    def validate_process(cls, value, values: dict) -> Union[None, int]:
+        if value is None:
+            return None
```
tim-vd-aardweg, line 94: return value

MRVermeulenDeltares: updated.

hydrolib/core/dimr/models.py
```diff
@@ -88,6 +88,78 @@ class FMComponent(Component):
 
     library: Literal["dflowfm"] = "dflowfm"
 
+    @validator("process", pre=True)
+    def validate_process(cls, value, values: dict) -> Union[None, int]:
```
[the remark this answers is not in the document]
MRVermeulenDeltares: Will do!

hydrolib/core/dimr/models.py
```diff
@@ -88,6 +88,78 @@ class FMComponent(Component):
...
+            return False
+
+        for value in split_values:
+            if not value.isdigit():
+                return False
+
+        return True
+
+    @classmethod
+    def _validate_process_as_int(cls, value: str, name: str) -> bool:
```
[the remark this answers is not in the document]
MRVermeulenDeltares: updated.

hydrolib/core/dimr/models.py
```diff
@@ -88,6 +88,78 @@ class FMComponent(Component):
 
     library: Literal["dflowfm"] = "dflowfm"
 
+    @validator("process", pre=True)
+    def validate_process(cls, value, values: dict) -> Union[None, int]:
+        if value is None:
+            return None
+
+        if isinstance(value, str) and cls._validate_process_as_str(value):
```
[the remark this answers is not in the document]
MRVermeulenDeltares: Will rename, will also do the same for the int validatation, this method is currently located under the DIMR.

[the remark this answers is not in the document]
MRVermeulenDeltares: also updated other private validation methods in the same style.

hydrolib/core/dimr/models.py
```diff
@@ -325,6 +397,66 @@ def _post_init_load(self) -> None:
             except NotImplementedError:
                 pass
 
+    def _save(self, save_settings: ModelSaveSettings) -> None:
+        dimr_as_dict = self.dict()
+        dimr_as_dict = self.update_dimr_dictonary_with_adjusted_fmcomponent_values(
```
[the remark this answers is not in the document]
MRVermeulenDeltares: updated.

hydrolib/core/dimr/models.py
```diff
@@ -325,6 +397,66 @@ def _post_init_load(self) -> None:
             except NotImplementedError:
                 pass
 
+    def _save(self, save_settings: ModelSaveSettings) -> None:
+        dimr_as_dict = self.dict()
```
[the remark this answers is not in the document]
MRVermeulenDeltares: Updated as per discussion, same thing for parse implemented.

hydrolib/core/dimr/models.py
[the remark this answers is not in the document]
MRVermeulenDeltares: added.

hydrolib/core/dimr/models.py
```diff
@@ -342,3 +438,73 @@ def _get_serializer(
...
+        if ":" in values:
+            semicolon_split_values = values.split(":")
+            start_value = int(semicolon_split_values[0])
+            end_value = int(semicolon_split_values[-1])
+            return end_value - start_value + 1
+
+        return len(values.split())
+
+    @classmethod
+    def _is_valid_process_string(cls, values: str) -> bool:
```
tim-vd-aardweg, line 477: Here we just get a single string value right? Not multitple strings?

hydrolib/core/dimr/models.py
```diff
@@ -342,3 +438,73 @@ def _get_serializer(
...
+
+        if cls._is_valid_process_string(process_value):
+            value_as_int = cls._parse_process(process_value)
+            component.update({"process": value_as_int})
+            data.update({"component": component})
+
+        return data
+
+    @classmethod
+    def _parse_process(cls, values: str) -> int:
```
tim-vd-aardweg, line 467: values is misleading, it's a single string value, right?

tests/test_model.py
```diff
@@ -522,3 +522,417 @@ def test_model_diskonlyfilemodel_field_is_constructed_correctly(
         assert relevant_field == input
     else:
         assert relevant_field.filepath == input
+
+
+class TestFmComponentProcessIntegrationWithDimr:
+    def get_fm_dimr_config_data(self, input_data_process):
+        dimr_config_data = f"""<?xml version="1.0" encoding="utf-8"?>
```
tim-vd-aardweg, line 529: no need to first put it into a variable, you can immediately return the string :)

tests/test_model.py
```diff
@@ -522,3 +522,417 @@ def test_model_diskonlyfilemodel_field_is_constructed_correctly(
...
+    <inputFile>test.mdu</inputFile>
+    <process>{input_data_process}</process>
+    <mpiCommunicator>DFM_COMM_DFMWORLD</mpiCommunicator>
+  </component>
+</dimrConfig>
+"""
+        return dimr_config_data
+
+    def get_fm_dimr_config_data_without_process(self):
+        dimr_config_data = """<?xml version="1.0" encoding="utf-8"?>
```
tim-vd-aardweg, line 548: Same here

tests/test_model.py
```diff
@@ -522,3 +522,417 @@ def test_model_diskonlyfilemodel_field_is_constructed_correctly(
...
+            temporary_dimr_config_file,
+            temporary_save_location,
+        ) = self.setup_temporary_files(tmp_path, dimr_config_data)
+
+        dimr_config = DIMR(filepath=temporary_dimr_config_file)
+        dimr_config.save(filepath=temporary_save_location)
+
+        assert dimr_config.component[0].process == expected_process
+
+    def test_dimr_with_fmcomponent_given_old_incorrect_style_for_setting_process(
```
tim-vd-aardweg, line 693: Please add to the name that old style should raise error so we dont have to look at the test to see what the expected behaviour is,

tests/test_model.py
```diff
@@ -522,3 +522,417 @@ def test_model_diskonlyfilemodel_field_is_constructed_correctly(
...
+        assert expected_message in str(error.value)
+
+    @pytest.mark.parametrize(
+        "input_process",
+        [
+            pytest.param(0),
+            pytest.param(-1),
+        ],
+    )
+    def test_fmcomponent_process_after_init_with_input_process_zero_throws_valueerror(
```
tim-vd-aardweg, line 924: Name of test is not correct. It does not only check process 0

tests/test_model.py
```diff
@@ -522,3 +522,417 @@ def test_model_diskonlyfilemodel_field_is_constructed_correctly(
...
+
+    @pytest.mark.parametrize(
+        "input_process",
+        [
+            pytest.param("lalala"),
+            pytest.param("123"),
+            pytest.param(3.5),
+        ],
+    )
+    def test_fmcomponent_process_after_init_with_incorrect_input_throws_valueerror(
```
tim-vd-aardweg, line 899: -3 and 0 are also incorrect values. Maybe use `invalid type` somwhere in the test instead?

tests/test_model.py
```diff
@@ -522,3 +522,417 @@ def test_model_diskonlyfilemodel_field_is_constructed_correctly(
...
+
+    @pytest.mark.parametrize(
+        "input_process",
+        [
+            pytest.param(0),
+            pytest.param(-1),
+        ],
+    )
+    def test_fmcomponent_process_after_init_with_input_process_zero_throws_valueerror(
+        self,
```
tim-vd-aardweg, line 925: These kind of unit tests should be moved to `\tests\dimr\test_dimr.py` instead of putting them at this high level integration tests file.

tests/test_model.py
```diff
@@ -522,3 +522,417 @@ def test_model_diskonlyfilemodel_field_is_constructed_correctly(
...
+
+    @pytest.mark.parametrize(
+        "input_process",
+        [
+            pytest.param("This is incorrect"),
+            pytest.param(1.4234),
+            pytest.param("1234556"),
+        ],
+    )
+    def test_dimr_with_fmcomponent_given_old_incorrect_style_for_setting_process2(
```
tim-vd-aardweg, line 717: Rather than adding a 2 at the end of the testcase name, specify the difference in the testname. E.g.:
`test_dimr_with_fmcomponent_given_invalid_type_...`

Also, why is 1234556 an invalid value? :D

hydrolib/core/dimr/models.py
MRVermeulenDeltares: will rename to process_value

tests/test_model.py
MRVermeulenDeltares: True, I put it like this during testing because I was debugging through it, will update.

tests/test_model.py
MRVermeulenDeltares: will update

tests/test_model.py
MRVermeulenDeltares: added _raise_valueerror

tests/test_model.py
MRVermeulenDeltares: I think something was forgotten when I made the next test and used the previous test as base :)
Good catch.

1234556 is invalid because a number is given as a string and that is not supported.
Number was randomly typed, will use other number for that.

tests/test_model.py
MRVermeulenDeltares: will update

tests/test_model.py
MRVermeulenDeltares: will add `zero_or_negative`

tests/test_model.py
MRVermeulenDeltares: Lets have a call where we want to place which test class of the two new test classes.

I had not found any other Component or FMcomponent test.
Thus I was unsure if I should have created a new test file or keep it on the same level as where DIMR() is tested

The `\tests\dimr\test_dimr.py` does not have any test that inits `DIMR`

Issue #559, opened 2023-07-11
Incorrect parsing of string that starts with "0Dxxxxx"


xldeltares:
**Describe the bug**
When reading a model that has storagenodes with id "0D05252", the string is parsed incorrectly as "0e05252"
A small test model is enclosed.
[dflowfm.zip](https://github.com/Deltares/HYDROLIB-core/files/12015691/dflowfm.zip)

**To Reproduce**
```
from hydrolib.core.dflowfm import FMModel
mdu = "FlowFM.mdu"
model = FMModel(mdu)
print(model.geometry.storagenodefile.storagenode[1].id)
```
The above steps returns: 0**e**05252

**Expected behavior**
The above steps should return" 0**D**05252

**Screenshots**
This can be verified in the nodeFile.ini:
![image](https://github.com/Deltares/HYDROLIB-core/assets/71022080/769a2031-e30c-4449-ad2c-68750ba5ce6b)

**Version info (please complete the following information):**
 - OS: Windows
 - Version 0.5.2

**Additional info**
In my use case of Eindhoven, this is rather critical because there are more than 1000 nodes are parsed wrongly after reading. See below figure, the red nodes are parsed wrongly.
The wrongly parsed string will be written again but wont match the network node ID anymore, causing run to fail.
![image](https://github.com/Deltares/HYDROLIB-core/assets/71022080/384c4caa-8ba6-407b-b259-b710ad331265)

priscavdsluis:
Another issue was reported via support: https://issuetracker.deltares.nl/browse/SOFTSUP-648

tim-vd-aardweg:
A side-effect of our fix is that you can now also set attributes that are floats using the Fortran scientific notation, e.g.:

```
from hydrolib.core.dflowfm import FMModel
mdu = "FlowFM.mdu"
model = FMModel(mdu)
model.geometry.storagenodefile.storagenode[1].area = "1d3")
```

tim-vd-aardweg:
@xldeltares Could you verify that this issue is fixed in branch: `bug/559_Incorrect-parsing-of-string-starting-with-0D`?

xldeltares:
@tim-vd-aardwegAfter testing the reported bug in the example model enclosed, I confirm that the issues has been fixed in ```bug/559_Incorrect-parsing-of-string-starting-with-0D```.  Glad that this has been fixed and thanks!

PR #625, opened 2024-04-17
bug: Incorrect parsing of string that starts with "0Dxxxxx"

MRVermeulenDeltares:
For reviewer:
I am not a fan of regex, but used it to keep in line with the already existing code.
Should we refactor that or leave it be?


MRVermeulenDeltares:
Closing PR, fixed in PR: #559

Issue #571, opened 2023-10-23
Add missing mdu keywords to FMModel class


veenstrajelmer:
**Is your feature request related to a problem? Please describe.**
Several mdu keywords are missing in the hydrolib core FMModel class. When reading a mdu file that is copied from the FM dia file, a lot of keywords are dropped when reading them. This should not happen. It is also not allowed to add these keywords to the FMModel, while at least part of them are documented (like `wrimap_windstress`)

MWE with attached mdu file:
```
import hydrolib.core.dflowfm as hcdfm
file_mdu = r"c:\Users\veenstra\Downloads\mdu_fromdia.mdu"
mdu = hcdfm.FMModel(file_mdu)
mdu.save(file_mdu.replace(".mdu","ignoreextra_hcdfm.mdu"))
# mdu.output.wrimap_windstress
```
The `allowextra` file was obtained by overwriting the `hcdfm.ini.models.INIBasedModel().Config.extra` parameter with `Extra.allow`. A diff between the `ignoreextra` and `allowextra` files show the mdu keywords missing in hydrolib-core.

**Describe the solution you'd like**
Inclusion of these keywords.

[mdu_fromdia.mdu.txt](https://github.com/Deltares/HYDROLIB-core/files/13067091/mdu_fromdia.mdu.txt)

[mdu_fromdia_hcdfm_allowextra.mdu.txt](https://github.com/Deltares/HYDROLIB-core/files/13067599/mdu_fromdia_hcdfm_allowextra.mdu.txt)

[mdu_fromdia_hcdfm_ignoreextra.mdu.txt](https://github.com/Deltares/HYDROLIB-core/files/13067523/mdu_fromdia_hcdfm_ignoreextra.mdu.txt)

tim-vd-aardweg:
I have performed a diff. The diff shows shows that there are 178 keywords that are in the `allowextra` file that are not in the `ignoreextra` file.  See [Keywords diff.txt](https://github.com/Deltares/HYDROLIB-core/files/14995625/Keywords.diff.txt)

- Should we add all these 178 keywords to the FlowFM model?
- Are they all described in the manual (asking since we need to know default values and what comment/description to use)?
@veenstrajelmer & @rhutten

tim-vd-aardweg:
The following keywords have been added:

|Section|Keyword|Type|Default value|
|-------------|-------------|-------------|-------------|
|geometry|gridenclosurefile|Optional[PolyFile]|None|
[...]

PR #628, opened 2024-04-18
#571: Add missing mdu keywords

tim-vd-aardweg:
Refs: #571


hydrolib/core/dflowfm/mdu/models.py
```diff
@@ -1263,6 +1400,75 @@ class Comments(INIBasedModel.Comments):
...
+            "Write boundary points to map file (1: yes, 0: no).", alias="wrimap_bnd"
+        )
+        wrishp_dambreak: Optional[str] = Field(
+            "Writing dambreaks to shape file (0=no, 1=yes).", alias="wrishp_dambreak"
+        )
+        wrimap_waterdepth_hu: Optional[str] = Field(
+            "Write water depths on u-points to map file (1: yes, 0: no).",
+            alias="wrimap_waterdepth_hu",
+        )
+        ncmapdataprecision: Optional[str] = Field(
```
MRVermeulenDeltares, line 1450: this is an int below, but this is talking about precision, should it be an int?

hydrolib/core/dflowfm/mdu/models.py
```diff
@@ -568,6 +704,7 @@ class Comments(INIBasedModel.Comments):
     pavbnd: float = Field(0.0, alias="pavBnd")
     pavini: float = Field(0.0, alias="pavIni")
     computedairdensity: bool = Field(False, alias="computedAirdensity")
+    stresstowind: float = Field(0.0, alias="stressToWind")
```
MRVermeulenDeltares, line 707: should the alias be with capital letters?

hydrolib/core/dflowfm/mdu/models.py
```diff
@@ -553,6 +685,10 @@ class Comments(INIBasedModel.Comments):
         computedairdensity: Optional[str] = Field(
             "Compute air density yes/no (), 1/0, default 0.", alias="computedAirdensity"
         )
+        stresstowind: Optional[str] = Field(
+            "Switch between Wind speed (=0) and wind stress (=1) approach for wind forcing.",
```
MRVermeulenDeltares, line 689: is this text correct? it sounds like a bool in this case, or am I interperting this incorrectly?

hydrolib/core/dflowfm/mdu/models.py
tim-vd-aardweg: I assume it is an `int`. The default value according to the very up-to-date manual is: `0` (an integer).

EDIT: I just checked the file Jelmer provided in the issue (the allow_extra file), there the description of this keyword more clearly indicates that it is indeed an integer with two valid values 0 and 1.

`Precision for NetCDF data in map files (0: double, 1: float)`

hydrolib/core/dflowfm/mdu/models.py
tim-vd-aardweg: According to the manual's default value, it's a float/double. Reading the description, it made more sense to make it an int with only two options: 0 or 1.

MRVermeulenDeltares (approved):

Issue #622, opened 2024-04-16
When there are unknown keywords in the mdu, an error should be given instead of them being silently dropped.


tim-vd-aardweg:
**Is your feature request related to a problem? Please describe.**
In #568 a request was made to give the user feedback on unsupported keywords instead of silently dropping them. Since the issue also requested other changes we decided to move this specific request into a separate issue.

**Describe the solution you'd like**
When we encounter an unknown mdu keyword, we should give an error to the user instead of silently dropping the keyword.

rhutten:
When you encounter unknown keywords. Please give the following warning: Unknown keywords are detected (in section of MDU file): list of unknown keywords. In case of unknown keywords are allowed (#630 - keywords), then the keywords will be kept in memory. (Note that these unknown keywords have no validation). Else the keywords will be dropped.

MRVermeulenDeltares:
I have updated the changes in #632.
I have added multiple tests to verify the functioning.
Besides that I tested this to verify how it looks to the actual user during loading or setting unknown keywords.
I used the following piece of testcode to get the following output:

```
from pydantic import Extra
from hydrolib.core.dflowfm.mdu.models import FMModel

print("")
print("first test reading from mdu file with unknown keys:")
print("")

model = FMModel("C:\\Users\\Verme_mn\\Downloads\\with_unknown_keywords.mdu")
print("")
print("second test setting unknown keys on the model object:")
print("")

model.geometry.tim = "123"
model.geometry.tim2 = "123"
model.geometry.tim3 = "123"
model.geometry.tim ="456"

model.volumetables.tim = "123"
model.volumetables.tim2 = "123"
model.volumetables.tim3 = "123"
model.volumetables.tim ="456"
```

When I set the configuration to allow unknown keywords I get the following output:
![Code_0SLFXQrPgP](https://github.com/Deltares/HYDROLIB-core/assets/102953403/c691abad-fc9c-4152-997c-04af9f98ed85)

When I set the configuration to not allow unknown keywords I get the following output:
![Code_8Cm5WVpimJ](https://github.com/Deltares/HYDROLIB-core/assets/102953403/3da09979-4968-4fa7-8af8-4cd577c3a508)

When the configuration is set to not allow unknown keywords, unknown keywords will throw an exception as before on setting with the model, e.g. `model.geometry.tim = "123"`

PR #629, opened 2024-04-19
feat: When there are unknown keywords in the mdu, a warning should be given instead of them being silently dropped. 

MRVermeulenDeltares:
#622


MRVermeulenDeltares:
continued in #632

PR #632, opened 2024-04-24
feat: Raise error for unknown keywords

MRVermeulenDeltares:
#622

- [x] Throw error when unknown keyword is located in the mdu
- [x] When multiple unknown keywords are located in the mdu, have the error contain multiple unknown keywords.


MRVermeulenDeltares:
After the update to throw an error on an unknown key multiple test started to fail.
I have done some reasearch to try and figure out why.
I came to a few conclusions:

1. Unknown keywords in the testfiles
2. Outdated keywords in the testfiles
3. Not defined keywords in the models (those keywords are in the manual)
4. Problems with subfiles being validatated and not containing properties in their classes that are defined in their files. (e.g. .bc files and crosssection .ini related files, might be more. )

This is what I have found in regard to the keywords in relation with the tests, I tried to tackle all keywords but I am not 100% sure I have them all.

tim-vd-aardweg:
In #634 I wanted to add this test:

```python
    def test_load_model_with_research_keywords_as_fmmodel_raises_error(self):
        input_mdu = (
                test_input_dir / "research" / "mdu_with_research_keywords_from_dia_file_2024.03_release.mdu"
        )

        with pytest.raises(ValueError) as e:
            _ = FMModel(filepath=input_mdu)

        expected_error = "Unknown keywords are detected in section"
        assert expected_error in str(e.value)
```

But it currently fails, since we have `Extra.ignore`. So I can't yet add that test. That test can only be added after this issue is implemented.

hydrolib/core/dflowfm/ini/models.py
```diff
@@ -55,6 +56,15 @@ class Config:
         extra = Extra.ignore
         arbitrary_types_allowed = False
 
+    def __init__(self, **data):
+        super().__init__(**data)
```
tim-vd-aardweg, line 60: Wouldn't it be more efficient to first check if there are unknown keywords and then call `super().__init__()`? It seems we have all the data we need to determine if there are unknown keywords. And if there are unknown keywords there is no reason to let pydantic work all its magic, since we are raising an error anyway?

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -622,3 +623,48 @@ def rename_keys_for_backwards_compatibility(
...
+        data: Dict[str, Any],
+        section_header: str,
+        fields: Dict[str, Any],
+        excluded_fields: Set,
+    ):
+        """
+        Notify the user of unknown keywords.
+
+        Args:
+            data (Dict[str, Any])   : Input data containing all set properties which are checked on unknown keywords.
```
tim-vd-aardweg, line 645: What are `set properties`? It's a dict, not a set? 😊

hydrolib/core/dflowfm/ini/models.py
```diff
@@ -55,6 +56,15 @@ class Config:
...
+        self._unknown_keyword_error_manager.raise_error_for_unknown_keywords(
+            data,
```
tim-vd-aardweg, line 62: It seems we are only using the keys of this dict in the `raise_error_for_unknown_keywords()` method. So just pass `data.keys()` and update the types expected by the method as you will now pass a list of strings instead of a dict.

tim-vd-aardweg: Nevermind, ignore this comment. It seems perfectly fine to just pass the dict :)

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -622,3 +623,48 @@ def rename_keys_for_backwards_compatibility(
...
+        if len(unknown_keywords) == 0:
+            return
+        
+        raise ValueError(f"Unknown keywords are detected in section: '{section_header}', '{unknown_keywords}'")
+
+    def _get_all_unknown_keywords(
+        self, data: Dict[str, Any], fields: Dict[str, Any], excluded_fields: Set
+    ) -> List[str]:
+        list_of_unknown_keywords = []
+        for name, _ in data.items():
```
tim-vd-aardweg, line 661: You can use: `for name in data`. There is no need to get the keys and values!

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -622,3 +623,48 @@ def rename_keys_for_backwards_compatibility(
...
+            if self._is_unknown_keyword(name, fields, excluded_fields):
+                list_of_unknown_keywords.append(name)
+
+        return list_of_unknown_keywords
+
+    def _is_unknown_keyword(
+        self, name: str, fields: Dict[str, Any], excluded_fields: Set
+    ):
+        return name not in fields and name not in excluded_fields
```
tim-vd-aardweg, line 670: In HYDROLIB-core we use pydantic to turn dictionaries into valid objects. Pydantic can map the key in the dictionary to the appropriate attribute by either looking at the field name or the field alias. This behaviour can be set in the config (`config.allow_population_by_field_name`). In our `BaseModel` we set `config.allow_population_by_field_name` to `True`. This means that we want to be able to map the data via either the name or the alias.

In your implementation, we say that a keyword is unknown if there is no attribute with the same name as the key. However, this will, for example, fail for the research keywords that are implemented in #642. All research keyword attribute names are prefixed with _research_. This will fail the mapping and they will be considered an unknown keyword. I therefore think that it is important to not only check the field names, but also their alias!

hydrolib/core/dflowfm/ini/models.py
```diff
@@ -55,6 +56,15 @@ class Config:
         extra = Extra.ignore
         arbitrary_types_allowed = False
 
+    def __init__(self, **data):
```
tim-vd-aardweg, line 59: Was there a reason we could not put this in a `root_validator`?

tim-vd-aardweg:
This doesn't work for unknown sections, right? Should you create a follow-up issue for this?

tests/dflowfm/test_mdu.py
```diff
@@ -388,3 +392,115 @@ def test_loading_fmmodel_model_with_both_ini_and_xyn_obsfiles(self):
             assert actual_point.x == expected_point.x
             assert actual_point.y == expected_point.y
             assert actual_point.name == expected_point.name
+
+    def test_mdu_unknown_keywords_loading_gives_message_for_missing_keyword(
```
tim-vd-aardweg, line 396:
```suggestion
    def test_mdu_unknown_keywords_loading_gives_message_for_unknown_keyword(
```

tests/dflowfm/test_mdu.py
```diff
@@ -388,3 +392,115 @@ def test_loading_fmmodel_model_with_both_ini_and_xyn_obsfiles(self):
...
+        autoStart             = 0             
+        pathsRelativeToParent = 0             
+        unknownkey            = something
+        """
+
+        tmp_mdu_path = tmp_path / "tmp.mdu"
+        tmp_mdu_path.write_text(tmp_mdu)
+
+        with patch(
+            "hydrolib.core.dflowfm.ini.models.INIBasedModel.Config"
```
tim-vd-aardweg, line 414: Why do we need this mock?

tests/dflowfm/test_mdu.py
```diff
@@ -388,3 +392,115 @@ def test_loading_fmmodel_model_with_both_ini_and_xyn_obsfiles(self):
...
+            section = "General"
+            name = "unknownkey"
+
+            assert (
+                f"Unknown keywords are detected in '{section}', these keywords will be dropped:"
+                in captured.out
+            )
+            assert name in captured.out
+
+    def test_mdu_unknown_keywords_loading_gives_message_for_missing_keyword2(
```
tim-vd-aardweg, line 430: Give it an appropriate name please

tests/dflowfm/test_mdu.py
```diff
@@ -388,3 +392,115 @@ def test_loading_fmmodel_model_with_both_ini_and_xyn_obsfiles(self):
...
+        unknownkey            = something
+        """
+
+        tmp_mdu_path = tmp_path / "tmp.mdu"
+        tmp_mdu_path.write_text(tmp_mdu)
+
+        FMModel(filepath=tmp_mdu_path)
+        captured = capsys.readouterr()
+
+        excluded_fields = ["comments", "datablock", "_header"]
```
tim-vd-aardweg, line 484: This test will still pass if we ever change the `_exclude_fields` of the model. So maybe use: `excluded_fields` = `model._exclude_fields` instead? Maybe even assert that the list is not empty?

tests/dflowfm/test_mdu.py
```diff
@@ -388,3 +392,115 @@ def test_loading_fmmodel_model_with_both_ini_and_xyn_obsfiles(self):
...
+        for excluded_field in excluded_fields:
+            assert excluded_field not in captured.out
+
+    def test_mdu_unknown_keywords_allow_extra_setting_field_gives_message(self, capsys):
```
tim-vd-aardweg, line 488: In my opinion there is no need to check all the different options for `Extra.<allow/forbid/ignore>`. The user should not edit the values we have set, since it's not a public setting. If they decide to change it anyway, we can't offer support/validation.

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -3,8 +3,9 @@
 from datetime import datetime
 from enum import Enum
 from operator import eq
-from typing import Any, Callable, Dict, List, Optional, Type
+from typing import Any, Callable, Dict, List, Optional, Set, Type
 
+from pydantic import Extra
```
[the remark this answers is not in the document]
MRVermeulenDeltares: will updated

hydrolib/core/dflowfm/ini/util.py
MRVermeulenDeltares: will updated

tests/dflowfm/test_mdu.py
MRVermeulenDeltares: tests werent updated yet it seems, will update them and rename them.

tests/dflowfm/test_mdu.py
MRVermeulenDeltares: I will remove/update the tests which test with extra, as you also stated in the comment below.

hydrolib/core/dflowfm/ini/models.py
MRVermeulenDeltares: I am not sure, but I think it had to do with the exclude fields not yet being initialized, thus always returning an empty list for it, will check this.

hydrolib/core/dflowfm/ini/models.py
MRVermeulenDeltares: I am not sure, but I think it had to do with the exclude fields not yet being initialized, thus always returning an empty list for it, will check this.

hydrolib/core/dflowfm/ini/models.py
MRVermeulenDeltares: I will ignore this comment :)

hydrolib/core/dflowfm/ini/util.py
MRVermeulenDeltares: how would you suggest doing the alias check?
Create a set of all the aliases of the fields and check if the name is not in that set?

tests/dflowfm/test_mdu.py
MRVermeulenDeltares: will update

hydrolib/core/dflowfm/ini/util.py
MRVermeulenDeltares: will adjust

hydrolib/core/dflowfm/ini/models.py
MRVermeulenDeltares: I checked it by moving the super().__init__() below the raise_error_for_unknown_keywords call. The new tests are still green, so I think we can shift it in that way.

hydrolib/core/dflowfm/ini/models.py
MRVermeulenDeltares: I think we can move it to the root validator, but the root validator is run after the init, so moving it might be in conflict with your statement below.

hydrolib/core/dflowfm/ini/models.py
tim-vd-aardweg: Nice! :)

hydrolib/core/dflowfm/ini/models.py
tim-vd-aardweg: Okay, let's leave it in the init.

hydrolib/core/dflowfm/ini/models.py
MRVermeulenDeltares: discussed, this appears to cause more test to failt, because the exclude_fields of structure does some checks which should be resolved by OO. This causes the check to crash when the exclude_fields is executed before init.

MRVermeulenDeltares: I will keep the init before the raise_error_for_unknown_keywords call.

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -622,3 +622,45 @@ def rename_keys_for_backwards_compatibility(
...
+    Detects unknown keys and manages the Error to the user.
+    """
+
+    def raise_error_for_unknown_keywords(
+        self,
+        data: Dict[str, Any],
+        section_header: str,
+        fields: Dict[str, ModelField],
+        excluded_fields: Set,
+    ):
```
tim-vd-aardweg, line 639:
```suggestion
    ) -> None:
```

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -622,3 +622,45 @@ def rename_keys_for_backwards_compatibility(
...
+    Error manager for unknown keys.
+    Detects unknown keys and manages the Error to the user.
+    """
+
+    def raise_error_for_unknown_keywords(
+        self,
+        data: Dict[str, Any],
+        section_header: str,
+        fields: Dict[str, ModelField],
+        excluded_fields: Set,
```
tim-vd-aardweg, line 638: Set[str]

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -622,3 +622,45 @@ def rename_keys_for_backwards_compatibility(
...
+        for name in data:
+            if self._is_unknown_keyword(name, fields, excluded_fields):
+                list_of_unknown_keywords.append(name)
+
+        return list_of_unknown_keywords
+
+    def _is_unknown_keyword(
+        self, name: str, fields: Dict[str, ModelField], excluded_fields: Set
+    ):
+        for model_field in fields.values():
```
tim-vd-aardweg, line 662: Nice 👍

hydrolib/core/dflowfm/ini/util.py
MRVermeulenDeltares: Agree, but... _exclude_fields from INIBaseModel returns a set, not a set[str], it is a set[str], but the typehint only suggests a set.

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -622,3 +622,45 @@ def rename_keys_for_backwards_compatibility(
...
+    """
+
+    def raise_error_for_unknown_keywords(
+        self,
+        data: Dict[str, Any],
+        section_header: str,
+        fields: Dict[str, ModelField],
+        excluded_fields: Set,
+    ):
+        unknown_keywords = self._get_all_unknown_keywords(data, fields, excluded_fields)
```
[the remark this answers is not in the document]
MRVermeulenDeltares: done

hydrolib/core/dflowfm/ini/util.py
MRVermeulenDeltares: done

hydrolib/core/dflowfm/ini/util.py
MRVermeulenDeltares: done

hydrolib/core/dflowfm/ini/util.py
MRVermeulenDeltares: thx

Issue #354, opened 2022-10-03
ExtModel reads some forcings multiple times


veenstrajelmer:
**Describe the bug**
When reading in an ExtModel, I see some duplicate forcings. 

The ext file contains:
```
[boundary]
quantity=dischargebnd
locationfile=../../../boundary_conditions/Lek.pli
forcingfile=../../../boundary_conditions/flow/rmm_rivdis_meas_20070101_20140101_MET.bc

[boundary]
quantity=dischargebnd
locationfile=../../../boundary_conditions/Waal.pli
forcingfile=../../../boundary_conditions/flow/rmm_rivdis_meas_20070101_20140101_MET.bc

[boundary]
quantity=dischargebnd
locationfile=../../../boundary_conditions/Maas.pli
forcingfile=../../../boundary_conditions/flow/rmm_rivdis_meas_20070101_20140101_MET.bc

[boundary]
quantity=salinitybnd
locationfile=../../../boundary_conditions/rmm_zeerand_v3.pli
forcingfile=../../../boundary_conditions/transport/rmm_zeesal_30years_v3.bc

[boundary]
quantity=temperaturebnd
locationfile=../../../boundary_conditions/rmm_zeerand_v3.pli
forcingfile=../../../boundary_conditions/transport/rmm_zeetemp_astronomic_v2.bc
```

ExtModel has five 'forcings' but the first three bc models all have three forcingobjects, which are the same for all three bc models:
```
boundary 1 of 5: ..\..\..\boundary_conditions\flow\rmm_rivdis_meas_20070101_20140101_MET.bc
forcing 1 of 3: Lek_0001 (timeseries) (dischargebnd)
forcing 2 of 3: Waal_0001 (timeseries) (dischargebnd)
forcing 3 of 3: Maas_0001 (timeseries) (dischargebnd)
boundary 2 of 5: ..\..\..\boundary_conditions\flow\rmm_rivdis_meas_20070101_20140101_MET.bc
forcing 1 of 3: Lek_0001 (timeseries) (dischargebnd)
forcing 2 of 3: Waal_0001 (timeseries) (dischargebnd)
forcing 3 of 3: Maas_0001 (timeseries) (dischargebnd)
boundary 3 of 5: ..\..\..\boundary_conditions\flow\rmm_rivdis_meas_20070101_20140101_MET.bc
forcing 1 of 3: Lek_0001 (timeseries) (dischargebnd)
[...]

PR #641, opened 2024-05-08
feat: reduce rereading cached files during init

MRVermeulenDeltares:
#354 

- [x] The existing caching should be expanded to also be used with initialization reading, instead of only newing objects.
- [x] The existing caching should be expanded to verify if a file has changed before returning the cache data.
- [x] When the file has changed a new cache of the file should be made.

### Note for reviewer
Im not fully sure how the @contextmanager functions.
It seems to create new instance for caching for each model which is initialized.

I could not setup a good test to test these steps:
Reading a file --> updating and changing the model it is read to --> saving the file --> reading the same file, changed data is read, since the file has been updated while the application is running.
Since there is not any other way to read the file as a user except for via the ctor.
The _load method could be used, but this is "private" and should not be used by the user.
This method also does not make use of the caching.

### DIFF
--- a/hydrolib/core/basemodel.py
+++ b/hydrolib/core/basemodel.py
@@ -431,6 +431,7 @@ class CachedFileModel:
         self._model = model
         self._checksum = checksum
 
+
 class FileModelCache:
     """
     FileModelCache provides a simple structure to register and retrieve FileModel
@@ -840,7 +841,7 @@ class FileModel(BaseModel, ABC):
                 if not context.is_content_changed(filepath):
                     cls._has_been_loaded_from_cache = True
                     return file_model
-            
+
             cls._has_been_loaded_from_cache = False
             return super().__new__(cls)
 
--- a/hydrolib/core/basemodel.py
+++ b/hydrolib/core/basemodel.py
@@ -403,22 +403,24 @@ class FilePathResolver:
         if relative_mode == ResolveRelativeMode.ToAnchor:
             self._anchors.pop()
 
-class CachedPathFileModelData():
 
-    _model : "FileModel"
-    _checksum : str
-    
+class CachedPathFileModelData:
+
+    _model: "FileModel"
+    _checksum: str
+
     @property
     def model(self) -> "FileModel":
         return self._model
-    
+
     @property
     def checksum(self) -> str:
         return self._checksum
 
-    def __init__(self, model : "FileModel", checksum : str) -> None:
+    def __init__(self, model: "FileModel", checksum: str) -> None:
         self._model = model
         self._checksum = checksum
+
 
 class FileModelCache:
     """
@@ -461,10 +463,10 @@ class FileModelCache:
             bool: Whether or not the cache is empty.
         """
         return not any(self._cache_dict)
-    
-    def exists(self, path : Path) -> bool:
+
+    def exists(self, path: Path) -> bool:
         """Whether or not the filepath is in the cache.
-        
+
         Args:
             path (Path): The path to verify if it is already added in the cache.
 
@@ -472,10 +474,10 @@ class FileModelCache:
             bool: Whether or not the path is in the cache.
         """
         return path in self._cache_dict
-    
-    def has_changed(self, path : Path) -> bool:
+
+    def has_changed(self, path: Path) -> bool:
         """Whether or not the file in the filepath has changed from the cache.
-        
+
         Args:
             path (Path): The path to verify verify against.
 
@@ -487,12 +489,13 @@ class FileModelCache:
         """
         if not self.exists(path):
             return True
-        
+
         checksum = self._get_checksum(path)
         return checksum != self._cache_dict.get(path).checksum
-    
-    def _get_checksum(self, path : Path) -> str:
+
+    def _get_checksum(self, path: Path) -> str:
         return FileChecksumCalculator.calculate_checksum(path)
+
 
 class ModelSaveSettings:
     """A class that holds the global settings for model saving."""
@@ -721,7 +724,7 @@ class FileLoadContext:
             file_path, self.load_settings.path_style
         )
         return Path(converted_file_path)
-    
+
     def is_content_changed(self, path: Path) -> bool:
         """Verify if the path is already known and if the content have changed.
 
@@ -731,7 +734,7 @@ class FileLoadContext:
         Args:
             path (Path): The relative path from which the model was loaded.
             model (FileModel): The loaded model.
-            
+
         Returns:
             True when the content from the path on the location has been changed.
             False when the content from the path is not priorly cached.
@@ -739,7 +742,6 @@ class FileLoadContext:
         """
         absolute_path = self._path_resolver.resolve(path)
         return self._cache.has_changed(absolute_path)
-        
 
 
 @contextmanager
@@ -878,7 +880,7 @@ class FileModel(BaseModel, ABC):
                 filepath = self._get_updated_file_path(filepath, loading_path)
 
             logger.info(f"Loading data from {filepath}")
-            
+
             data = context.retrieve_model(filepath)
             if context.is_content_changed(filepath):
                 data = self._load(loading_path)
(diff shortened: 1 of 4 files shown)


hydrolib/core/utils.py
```diff
@@ -300,3 +301,34 @@ def _from_windows_to_posix_path(cls, windows_path: Path) -> str:
...
+
+        Args:
+            filepath (Path): The filepath to the file for which the checksum will be calculated.
+
+        Returns:
+            [Optional[str]]:
+                The checksum of the file.
+                When the filepath doesn't exist or the filepath isn't a file, None.
+        """
+        if not filepath.exists() or not filepath.is_file():
```
tim-vd-aardweg, line 323: Is there a reason you chose to return `None` instead of raising an error?

hydrolib/core/utils.py
```diff
@@ -300,3 +301,34 @@ def _from_windows_to_posix_path(cls, windows_path: Path) -> str:
...
+            return None
+
+        return FileChecksumCalculator._calculate_sha256_checksum(filepath)
+
+    @staticmethod
+    def _calculate_sha256_checksum(filepath: Path) -> str:
```
tim-vd-aardweg, line 329: Shouldn't a MD5 checksum be enough to check if there are differences? MD5 is faster and we do not need the cryptographic security SHA256 offers :p

hydrolib/core/basemodel.py
```diff
@@ -403,6 +404,29 @@ def pop_last_parent(self) -> None:
             self._anchors.pop()
 
 
+class CachedPathFileModelData:
+    """
+    CachedPathFileModelData provides a simple structure to keep the Filemodel and checksum together.
+    """
+
+    _model: "FileModel"
```
tim-vd-aardweg, line 412: I dont think these attribute declarations are required

hydrolib/core/basemodel.py
```diff
@@ -403,6 +404,29 @@ def pop_last_parent(self) -> None:
...
+    def model(self) -> "FileModel":
+        """FileModel."""
+        return self._model
+
+    @property
+    def checksum(self) -> str:
+        """ "Checksum of the file the filemodel is based on."""
+        return self._checksum
+
+    def __init__(self, model: "FileModel", checksum: str) -> None:
```
tim-vd-aardweg, line 425: Missing documentation

hydrolib/core/basemodel.py
```diff
@@ -403,6 +404,29 @@ def pop_last_parent(self) -> None:
             self._anchors.pop()
 
 
+class CachedPathFileModelData:
```
tim-vd-aardweg, line 407: This name confuses me. Maybe just use `CachedFileModel`.

hydrolib/core/basemodel.py
```diff
@@ -422,7 +446,10 @@ def retrieve_model(self, path: Path) -> Optional["FileModel"]:
                 The FileModel associated with the Path if it has been registered
                 before, otherwise None.
         """
-        return self._cache_dict.get(path, None)
+        file_model = self._cache_dict.get(path, None)
+        if file_model is None:
```
tim-vd-aardweg, line 450: These two lines are redundant. If the given `path` is not in the cache, `file_model` is `None` and will be returned on line 452.

hydrolib/core/basemodel.py
```diff
@@ -441,6 +469,38 @@ def is_empty(self) -> bool:
         """
         return not any(self._cache_dict)
```
tim-vd-aardweg, line 470: This should also work: `return not self._cache_dict`. Knowing that, this function may not be needed since the statement is so simple.

hydrolib/core/basemodel.py
```diff
@@ -441,6 +469,38 @@ def is_empty(self) -> bool:
...
 
+    def exists(self, path: Path) -> bool:
```
tim-vd-aardweg, line 472: Can be made private.

hydrolib/core/basemodel.py
```diff
@@ -441,6 +469,38 @@ def is_empty(self) -> bool:
...
+            True when the file does not exist in caching reference.
+            False when the file has not changed.
+        """
+        if not self.exists(path):
+            return True
+
+        checksum = self._get_checksum(path)
+        return checksum != self._cache_dict.get(path).checksum
+
+    def _get_checksum(self, path: Path) -> str:
```
tim-vd-aardweg, line 501: If you decide to make `calculate_checksum()` return an optional string, then also make this return value an `Optional[str]`

hydrolib/core/basemodel.py
```diff
@@ -808,13 +886,13 @@ def __init__(
...
             logger.info(f"Loading data from {filepath}")
 
-            data = self._load(loading_path)
-            context.register_model(filepath, self)
-            data["filepath"] = filepath
+            data = context.retrieve_model(filepath)
+            if context.is_content_changed(filepath):
+                data = self._load(loading_path)
+                context.register_model(filepath, self)
+                data["filepath"] = filepath
```
tim-vd-aardweg, line 893: Shouldn't the filepath be set regardless of whether the data has been cached or not? Another question: If we find data in the cache, do we still need to do all the other steps below? If a file is referenced multiple times, do we need separate instances of models, or should they refer to the same instance?

If `context.retrieve_model(filepath)` returns something (i.e. the filepath is in the cache), `data` is a `FileModel`. If the model is not cached, `data` will be a dictionary.

I think we should discuss this.

hydrolib/core/utils.py
tim-vd-aardweg: Discussed. We will change it to MD5.

hydrolib/core/basemodel.py
MRVermeulenDeltares: so remove them?

hydrolib/core/basemodel.py
MRVermeulenDeltares: will do

hydrolib/core/basemodel.py
MRVermeulenDeltares: will update

hydrolib/core/basemodel.py
MRVermeulenDeltares: no, since line 452 is calling the model from the cache, the CachedFileModel, will rename it to cached_file_model

hydrolib/core/basemodel.py
MRVermeulenDeltares: i think this function add for readability wise.

hydrolib/core/basemodel.py
MRVermeulenDeltares: will do

hydrolib/core/basemodel.py
MRVermeulenDeltares: will do

hydrolib/core/utils.py
MRVermeulenDeltares: it has to do with the weird files it tries to access on each run, which we also encountered when we were pair programming on this.

hydrolib/core/basemodel.py
MRVermeulenDeltares: done

hydrolib/core/basemodel.py
MRVermeulenDeltares: exist

hydrolib/core/basemodel.py
MRVermeulenDeltares: we worked this out together.
I now also added rereading when file has changed back in, in the new.

hydrolib/core/utils.py
MRVermeulenDeltares: This has become a security risk in sonarcloud

tim-vd-aardweg: Marked it safe in sonarcloud, as it poses no security risk in our context.

hydrolib/core/basemodel.py
tim-vd-aardweg: If possible, yes.

hydrolib/core/basemodel.py
tim-vd-aardweg: Slightly unconventional, but personal preference in the end ^^

tim-vd-aardweg (approved):

PR #645, opened 2024-06-03
mesh_2d_edge_x and mesh_2d_edge_y to writer

DanielTollenaar:
Improving `UgridWriter` and other to be able to read net-files as UGRID using `MDAL` (e.g. within QGIS). 

**Added**:
- Enhanced `UgridWriter` by adding `mesh2d_edge_x` and `mesh2d_edge_y` reading it from `network._mesh2d` (`meshkernel.Mesh2d`)


veenstrajelmer:
@DanielTollenaar Is this still relevant? Why is the PR marked as a draft?

DanielTollenaar:
@veenstrajelmer, For DHYDRO users I do think it is relevant for the reason mentioned in this draft (2D meshes cannot be read by MDAL due to incompleteness).

It's draft as I saw other imperfections (e.g. projection-info) I could add.

It's left as I assumed this repo isn't actively maintained by it's owners. As a user it doesn't make much sense to contribute if PR's aren't reviewed nor merged.

veenstrajelmer:
Ok, good to know. The network part of the code is being refactored in several steps. Metadata like crs is also valuable indeed, but this also requires an awareness of crs in hydrolib-core. In the future, we will be looking into external packages like xugrid or ugridpy to outsource the netcdf writing toward. However, I think for the short term this PR is already a good addition and we will consider merging it today.

.gitignore
```diff
@@ -138,3 +138,6 @@ tests/data/output
 
 # PyCharm
 .idea/
+
+# Put your own experiments here
+brew/*
```
veenstrajelmer, line 143: Please remove this addition to minimize the diff, it does not seem to add value and it will start our checks/ci again which will hopefully be green.

veenstrajelmer (changes requested):
@DanielTollenaar the checks are failing, but we cannot rerun them on your fork-PR somehow. Also, we cannot commit to it, this would also kick off the checks again. A simple solution is to copy your edits to a new PR, but we would like to leave the credits with you. Also I have one request to minimize the diff. Would you have time to do this?

DanielTollenaar:
> Ok, good to know. The network part of the code is being refactored in several steps. Metadata like crs is also valuable indeed, but this also requires an awareness of crs in hydrolib-core. In the future, we will be looking into external packages like xugrid or ugridpy to outsource the netcdf writing toward. However, I think for the short term this PR is already a good addition and we will consider merging it today.

Agreed; refactoring seems absolutely necessary and I prefer the setup of XUGRID https://deltares.github.io/xugrid/ as it's more intuitive (build on xarray).

It may be good to have a look at this issue: https://github.com/Deltares/MeshKernel/issues/326 and work-around it in the UGridWriter as long as it isn't fixed in MeshKernel. Now writing will cause a crash if you manipulate the mesh before.

veenstrajelmer:
Thanks for pointing to https://github.com/Deltares/MeshKernel/issues/326, I was unaware of it since I only monitor the [meshkernelpy](https://github.com/Deltares/MeshKernelPy/issues) issues every now and then, the python API of meshkernel that you are using there. I can understand the difference is a bit arbitrary for externals. Either way, I commented the current status there and linked some related issues. This is indeed important to fix, but that will take a bit more time.

.gitignore
DanielTollenaar: done

veenstrajelmer:
Test is still failing. Could you update your branch with main? I expect that your changes should not cause the fm_container run to fail, but I also do not see the ci run being kicked off, so something seems broken. If updating your branch to main does not work, we will merge the changes in a parallel branch.

veenstrajelmer (changes requested):
Test is still failing. Could you update your branch with main? I expect that your changes should not cause the fm_container run to fail, but I also do not see the ci run being kicked off, so something seems broken. If updating your branch to main does not work, we will merge the changes in a parallel branch.

DanielTollenaar:
> Test is still failing. Could you update your branch with main? I expect that your changes should not cause the fm_container run to fail, but I also do not see the ci run being kicked off, so something seems broken. If updating your branch to main does not work, we will merge the changes in a parallel branch.

Done, but the reason for failing suggests something with a missing username and password, doubt merging from upstream will fix that. Feel free to just copy-paste the code in a new branch in your repo. Don't care about credits ;-):

image

veenstrajelmer:
Yes, I noticed that also, but wanted to make sure it was not because of the branch being outdated. I think we have to look into PR's from forks in the future, since this seems to be the issue. All tests (both ci and fm_container) are green on other recent branches. We will discuss how to approach this. Thanks for the contribution either way!

tim-vd-aardweg (approved):

PR #647, opened 2024-06-06
feat: external forcings converter

### LINKED ISSUE #621: Converter for old external forcings file
D-Flow FM has an "old external forcings file" ([ExtOldModel](https://deltares.github.io/HYDROLIB-core/0.5.2/reference/extold/#hydrolib.core.dflowfm.extold.models.ExtOldModel) in HYDROLIB-core), which is about to be phased out and replaced by blocks in:
* new external forcings file ([ExtModel](https://deltares.github.io/HYDROLIB-core/0.5.2/reference/ext/#hydrolib.core.dflowfm.ext.models.ExtModel) in HYDROLIB-core)
* structures file ([StructureModel](https://deltares.github.io/HYDROLIB-core/0.5.2/reference/structure/#hydrolib.core.dflowfm.structure.models.StructureModel) in HYDROLIB-core))
* initial fields and parameters file ([IniFieldModel](https://deltares.github.io/HYDROLIB-core/0.5.2/reference/inifield/#hydrolib.core.dflowfm.inifield.models.IniFieldModel) in HYDROLIB-core))

We would like a converter that takes an *.mdu file as input, scans for any old ext file, and exports a new MDU file, with underlying new attribute files as listed above.

A script, both callable as a function, as well as from the commandline.

**Additional context**
See D-Flow FM kernel issue UNST-7872 for the actual migration epic.

arthurvd:
Closes #621


hydrolib/core/dflowfm/mdu/legacy.py
arthurvd, line 1: @priscavdsluis / @tim-vd-aardweg : what do you guys think about this extra "legacy.py" inside the MDU directory?
As long as we don't have versioned support for several old/newer formats of some of our filetypes, this was the solution I came up with.

tim-vd-aardweg: @arthurvd / @priscavdsluis 
I am not entirely sure to be honest. My first question is: Do we  want tools in HYDROLIB-core, or do we create a separate package for tools that use HYDROLIB-core?

But let's assume we do want to add tools to the HYDROLIB-core package. I think my next question is: Should we implement this 'legacy' model into hydrolib, or should it be defined in the tool that uses it? Also, the change you request touches on a broader topic that should be discussed: How do we handle old/deprecated files/sections/keywords. Your requested change is due to the fact that we cannot handle older/unexpected file formats. And your change fixes things for your specific use case.

Having said all that, my thoughts are as follows:
- We will need to generically support older/unexpected formats. That way, we don't need your specific legacy model.
- Since that probably won't happen in the short term, we need a different solution for your use case. My preference would be to add the LegacyModel to the specific tool, so we do not pollute the main HYDROLIB-core package with quick fixes for specific tools.

MRVermeulenDeltares: Hi @arthurvd , I know you didn't ask my opinion but I still want to ask some questions regarding this.
Previously I remember a presentation was given with a picture showing hydrolib-core and hydrolib.
And how it should be used by tools.
See https://www.deltares.nl/expertise/projecten/hydrolib-gezamenlijke-automatisering-op-watersysteemmodellen
image


Should the converter not be a tool build on top of hydrolib-core?
Especially if the support for the old file format is dropped and no longer backwards compatible supported by the kernels.
Should hydrolib-core still support the old format, or only the new format from version x?

RStolkerDeltares:
@arthurvd I'm trying to run the converter script but it's failing the following error message below.
I did install the poetry environment ask described in the docs.

```
poetry run python main_converter.py
Traceback (most recent call last):
  File "main_converter.py", line 35, in <module>
    from .converter_factory import ConverterFactory
ImportError: attempted relative import with no known parent package
```

MAfarrag (changes requested):
can you fix the sonar cloud issues @jeroen-deltares

MAfarrag (changes requested):

Issue #655, opened 2024-07-03
Update MeshKernel to 4.3.0


tim-vd-aardweg:
**What is the task?**
Update MeshKernel from 4.1.0 to the latest release 4.3.0.

tim-vd-aardweg:
@rhutten 
Discussed this with @veenstrajelmer. We came to the conclusion that no action is required. Currently, we specify that we want a _minimum_ version of 4.1. However, in our current environment the MeshKernel is solved to version 4.3. Since there was no new functionality in 4.2 or 4.3 that Hydrolib-core uses, we did not see any reason to change the constraints on the dependency.

tldr: we leave the minimum version that can be used on 4.1. However, we currently use 4.3.

PR #656, opened 2024-07-03
chore: Update MeshKernel to 4.3.0.

tim-vd-aardweg:
refs: #655


tim-vd-aardweg:
We will not do this. See https://github.com/Deltares/HYDROLIB-core/issues/655#issuecomment-2209257301

PR #674, opened 2024-07-05
updated with suggestions of @KReef

veenstrajelmer:


hydrolib/core/dflowfm/net/models.py
```diff
@@ -590,12 +590,29 @@ class Link1d2d(BaseModel):
...
+        return link1d2d_arr
+
+    @property
+    def link1d2d_contact_type(self) -> np.ndarray[np.int32]:
+        contacts = self.meshkernel.contacts_get()
+        link1d2d_contact_type_arr = np.full(contacts.mesh1d_indices.size, 3)
+        return link1d2d_contact_type_arr
+
+    @property
+    def link1d2d_id(self) -> np.ndarray[object]:
```
tim-vd-aardweg, line 608: `link1d2d_id` and `link1d2d_long_name` do exactly the same, or am I missing something?

veenstrajelmer: Yes, that is true, that was also the case in the old code and I merely refactored it: https://github.com/Deltares/HYDROLIB-core/pull/674/files/2b7993a7a5bc4bad7fe495fe66cf13dc4b173869#diff-74029ac7f1b5ce72a91273e7a360d6da01995b163310293e671849e2d84de7dfL643-L650

tim-vd-aardweg: Okay, let's leave it like this for now.

priscavdsluis (approved):

PR #685, opened 2024-07-09
684 add test for ugrid compliance of netcdf network files

veenstrajelmer:


tests/dflowfm/test_net.py
```diff
@@ -765,3 +766,62 @@ def test_network_is_geographic():
...
+                if len(candidates) == 0:
+                    raise KeyError(
+                        f"the following variables are specified for UGRID {name}: "
+                        f'"{attrs[name]}", but they are not present in the dataset'
+                    )
+
+
+def test_network_netcdf_consistency(tmp_path):
+    from hydrolib.core.dflowfm import NetworkModel
+    import xarray as xr
```
tim-vd-aardweg, line 795: xarray is already imported at the top of the file? You can also add the import of the NetworkModel at the top of the test file (if it is not there yet)

Issue #723, opened 2024-12-17
Implement ability to pass meshkernel refinement parameters


RuudHurkmans:
In net/models.py a function is defined to pass refinement parameter to Meshkernel:
 

 def refine(self, polygon: mk.GeometryList, level: int, min_edge_size: float = 10.0):
        """Refine the mesh within a polygon, by a number of steps (level)   
       
     parameters = mk.MeshRefinementParameters(
            refine_intersected=True,
            use_mass_center_when_refining=False,
            min_edge_size=min_edge_size,
            refinement_type=1,
            connect_hanging_nodes=True,
            account_for_samples_outside_face=False,
            max_refinement_iterations=level,
        )

However, it's not possible to pass most of the parameters from an external function. Is it possible to add 'refine_intersected', "use_mass_center_when_refining", "refinement_type","connect_hanging_nodes" and "account_for_samples_outside_face" as arguments  to two functions:

Mesh2d.refine() and Network.mesh2d_refine_mesh()? This enables setting the parameters from hydrolib. Especially "use_mass_center_when_refining" gives much better results.

rhutten:
Issue and PR can be closed after wrapping-up #776. See #776 for more information.

PR #738, opened 2025-02-12
implemented ability to pass on meshkernel parameters from hydrolib 

RuudHurkmans:
Extended refinement functions to allow passing of meshkernel parameters and removed checks for clipping (functionality they say is not implemented does in fact work).

In the process the default values that were hardcoded in hydrolib core were changed to the same values as are used in meshkernel (they make more sense), hense the asserts in the refinement test failed, as the number of cells changed slightly. They test has been updated.


MAfarrag:
solved in #723

PR #792, opened 2025-04-01
Fix: quantity file discovery with _ after stem.

jeroen-deltares:
Added underscore after file stem for glob of files.

### DIFF
--- a/hydrolib/tools/extforce_convert/converters.py
+++ b/hydrolib/tools/extforce_convert/converters.py
@@ -281,7 +281,9 @@ class BoundaryConditionConverter(BaseConverter):
             )
 
         forcings_list = []
-        tim_files = list(self.root_dir.glob(f"{poly_line.filepath.stem}_[0-9][0-9][0-9][0-9].tim"))
+        tim_files = list(
+            self.root_dir.glob(f"{poly_line.filepath.stem}_[0-9][0-9][0-9][0-9].tim")
+        )
         if len(tim_files) > 1:
             time_series_list = self.convert_tim_to_bc(
                 tim_files, time_unit, quantity=quantity, label=label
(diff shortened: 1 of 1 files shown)


hydrolib/tools/extforce_convert/converters.py
```diff
@@ -281,15 +281,15 @@ def convert(self, forcing: ExtOldForcing, time_unit: str) -> Boundary:
             )
 
         forcings_list = []
-        tim_files = list(self.root_dir.glob(f"{poly_line.filepath.stem}*.tim"))
+        tim_files = list(self.root_dir.glob(f"{poly_line.filepath.stem}_*.tim"))
```
arthurvd: This is not safe against, for example:
right_mister.pli (with right_mister_0001.tim) and right_mister_sal.pli (with right_mister_sal_0001.tim).
I think the change below is safer:
```suggestion
        tim_files = list(self.root_dir.glob(f"{poly_line.filepath.stem}_[0-9][0-9][0-9][0-9].tim"))
```

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -281,15 +281,15 @@ def convert(self, forcing: ExtOldForcing, time_unit: str) -> Boundary:
...
         if len(tim_files) > 1:
             time_series_list = self.convert_tim_to_bc(
                 tim_files, time_unit, quantity=quantity, label=label
             )
             forcings_list.extend(time_series_list)
 
         # check t3d files
-        t3d_files = list(self.root_dir.glob(f"{poly_line.filepath.stem}*.t3d"))
+        t3d_files = list(self.root_dir.glob(f"{poly_line.filepath.stem}_*.t3d"))
```
arthurvd: same here

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -303,7 +303,7 @@ def convert(self, forcing: ExtOldForcing, time_unit: str) -> Boundary:
             forcings_list.extend(t3d_forcing_list)
 
         # check cmp files
-        cmp_files = list(self.root_dir.glob(f"{poly_line.filepath.stem}*.cmp"))
+        cmp_files = list(self.root_dir.glob(f"{poly_line.filepath.stem}_*.cmp"))
```
arthurvd: same here

arthurvd (changes requested):

hydrolib/tools/extforce_convert/converters.py
jeroen-deltares: applied

hydrolib/tools/extforce_convert/converters.py
jeroen-deltares: applied

arthurvd (approved):

PR #1135, opened 2026-06-26
fix(timeseries): add validation to ensure datablock columns match quantity unit pairs

### LINKED ISSUE #305: Validation upon initialisation of TimeSeries is not strict enough
**Describe the bug**
Initialisation of TimeSeries does not fail when supplying incorrect number of QuantityUnitPairs for the amount of columns.

**To Reproduce**
```
from hydrolib.core.io.bc.models import QuantityUnitPair, TimeInterpolation, TimeSeries
steric = TimeSeries(
    name="east2_0001",
    quantityunitpair=[#QuantityUnitPair(quantity="time", unit="seconds since 2022-01-01 00:00:00 +00:00"),
                      QuantityUnitPair(quantity="waterlevel", unit="m")],
    timeinterpolation=TimeInterpolation.linear,
    datablock=[[0, 1.1], [3600, 2.3], [7200, 3.5], [10800, 2.4], [86400, -0.123]],
)
```

**Expected behavior**
Since there is only QuantityUnitPair supplied and there are two columns, the initialization of TimeSeries should fail I think.

**Screenshots**

 - OS: Windows, Anaconda, Spyder
 - Version: 0.3.0

**Additional context**

riordache12:
# Description
- Added a ForcingBase validator that compares the amount of `QuantityUnitPair` with the amount of columns. Also takes into consideration the possiblity that the `QuantityUnitPair` instances could be inside `VectorQuantityUnitPairs`

- Fixes # (issue)
#305 

Check relevant points.

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- Added tests that check if an error is raised when there are more `QuantityUnitPair` than columns
- Added tests that check if an error is raised when there are fewere `QuantityUnitPair` than columns
- Added tests that check if the validator passes when there is an equal amount of `QuantityUnitPair` and columns

# Checklist:
Prepare items below using:

- [ ] updated the lock file.


riordache12:
@veenstrajelmer - Could you please review the functionality. My expectation is that any inconsistency between number of QuantityUnitPair and number of columns should be picked up by the validator. I have also written tests to check that. Older tests are failing since there was no validator to pick up the configurations, but I want to make sure you approve it before I fix those tests

veenstrajelmer (dismissed):
Thanks, this looks good! However, I realized this is not only an issue for `TimeSeries` but for all related classes. I've updated the issue description with an example and the overview of classes: https://github.com/Deltares/HYDROLIB-core/issues/305#issue-1340572258.

About the tests: I've tested `test_time_series_discharge_case` and indeed the input is incorrect, so it makes sense that this new error is raised. The test input can indeed be updated.

riordache12:
> Thanks, this looks good! However, I realized this is not only an issue for `TimeSeries` but for all related classes. I've updated the issue description with an example and the overview of classes: [#305 (comment)](https://github.com/Deltares/HYDROLIB-core/issues/305#issue-1340572258).
> 
> About the tests: I've tested `test_time_series_discharge_case` and indeed the input is incorrect, so it makes sense that this new error is raised. The test input can indeed be updated.

Oh I see, so that means I should move the validator the `ForcingBase` so all its children classes can inherit it.

hydrolib/core/dflowfm/bc/models.py
```diff
@@ -347,6 +347,33 @@ def __repr__(self) -> str:
         representable = type(self).model_construct(**data)
         return str(representable)
 
+    @model_validator(mode="after")
```
MAfarrag, line 350: a field validator is better here since you want to validate only the datablock field

hydrolib/core/dflowfm/bc/models.py
```diff
@@ -347,6 +347,33 @@ def __repr__(self) -> str:
...
+    def _validate_datablock_columns_match_quantityunitpairs(self) -> "ForcingBase":
+        """Validates that the number of columns in the datablock matches the total number of quantity unit pairs.
+
+        Raises:
+            ValueError: When the number of datablock columns does not match the number of quantity unit pairs.
+
+        Returns:
+            ForcingBase: The validated model instance.
+        """
+        if not self.datablock:
```
MAfarrag, line 360: when you use field validator you won't need this  
```python
if not self.datablock:
...
```

hydrolib/core/dflowfm/bc/models.py
```diff
@@ -347,6 +347,33 @@ def __repr__(self) -> str:
         representable = type(self).model_construct(**data)
         return str(representable)
 
+    @model_validator(mode="after")
+    def _validate_datablock_columns_match_quantityunitpairs(self) -> "ForcingBase":
```
MAfarrag, line 351: why did you add the validator in this `ForcingBase` class ? the datablock field exist in the parent `DataBlockINIBasedModel` class, so move the validator there.

MAfarrag (changes requested):

hydrolib/core/dflowfm/bc/models.py
riordache12: But `quantityunitpair` field exists under `ForcingBase` class and I would also need to import `VectorQuantityUnitPairs` inside that file which I am pretty sure would give me a circular import error.

hydrolib/core/dflowfm/bc/models.py
riordache12: Correct me if I am wrong, but we're actually performing a cross-field check here, right? Both `datablock` and `quantityunitpair` are required for this validation to succeed? I've tried implementing your suggestion but ended up not being able to access `quantityunitpair`, which according to Copilot is due to field order dependency.

tests/dflowfm/ext/test_ext.py
```diff
@@ -517,22 +517,22 @@ def test_save_preserves_inline_comment_text(self, tmp_path: Path):
...
+        "datablock": [[1, 2], [1.1234, 2.3456]],
     },
     {
         "name": "user_defined_name_1",
         "function": "timeseries",
         "timeinterpolation": "linear",
-        "quantity": ["time", "salinitydelta"],
-        "unit": ["minutes since 2015-01-01 00:00:00", "ppt"],
+        "quantity": ["time", "salinitydelta", "ux", "uy", "uz"],
+        "unit": ["minutes since 2015-01-01 00:00:00", "ppt", "m", "m", "m"],
```
MAfarrag: can you please change the unit `"m"`  to `"m s-1"` ?

MAfarrag (approved):

Issue #838, opened 2025-05-27
extforce-convert unnecessarily rewrites bc-files


veenstrajelmer:
**Describe the bug**
When running extforce-convert, the bc-files are rewritten. This is unneccesary and time consuming for actual (larger) models. Also, most modellers prefer tools not to mess with their data if not necessary.

**To Reproduce**
1. Copy model input from n:\Deltabox\Postbox\Farrag, Mostafa\fromJelmer, this was generated by the [dfm_tools modelbuilder notebook](https://github.com/Deltares/dfm_tools/blob/main/docs/notebooks/modelbuilder_example.ipynb)
2. both original and converted model is already provided, the converted model is generated with `extforce-convert --mdufile Vietnam.mdu`
3. the date of the bc-files has changed

**Expected behavior**
The bc file is not changed (besides the hydrolib-core version in the comments), so these files should be left untouched.

**Version info (please complete the following information):**
 - OS: Windows
 - Version 0.9.0

PR #1145, opened 2026-07-08
fix(models): .bc files are cleared

### LINKED ISSUE #1143: bc files are overwritten with empty contents
**Describe the bug**
.bc files are rewritten with an empty content since the 1.0.0b5 release

**To Reproduce**
call `extforce-convert` on the Vietnam model

**What happens**
The .bc files are cleared except for the header

**Expected behavior**
.bc files not being touched and just carrying over the same information

**Problem**
What actually happens here is that the .ext file gets loaded with `recurse=False` and the field validator `validate_forcingfile` calls ForcingModel(filepath=data) directly.

Then, inside `FileModel` class, there is a classmethod called `_should_load_model` which checks whether `context.load_settings.recurse or context.cache_is_empty()`. Since recurse is False that also returns False which meanse the .bc file is never read. The ForcingModel is initialized with an empty forcing=[] list. When the `save()` happens, the serialization overwrites the .bc file with an empty dump.

 - OS: Windows
 - Version 1.0.0/1.0.1

**Additional context**
- worked fine in tag [pre-merge/main-before-release-1.0.0](https://github.com/Deltares/HYDROLIB-core/releases/tag/pre-merge%2Fmain-before-release-1.0.0)
- tags 1.0.0 and 1.0.0b5 are available in https://github.com/Deltares/HYDROLIB-core/tags, but cannot be checked out and are also not printed with `git tag -l`
- is broken in 1.0.1

riordache12:
# Description

ATTENTION! Go through the story below and then check the implementation in #1149. That also fixes the issues described here. Ignore the implementation performed here as it only takes care of a certain usecase.

Read the Problem Description as I understand it in the issue linked to it #1143. 

Fixes #1143
Now, the first solution I thought of is to turn the `recurse` to True on line 135 inside `main_converter.py`. However that was change due to reports of being slow in #838, and it was not a real problem until last February when the validator of a forcing file was introduced

I looked inside the validator and realized it never uses the neat `resolve_file_model()` function defined in `utils.py`. But that function will always identify the file as a `DiskOnlyFileModel`. Which is never serialized so never rewrites the .bc file. I have considere extending the validator to accept the `DiskOnlyFileModel`, but I don't think that's the right way to go.

I ended up doing the solution implemented but I want to discuss it with you, @MAfarrag and see if you have a better idea.

Check relevant points.

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- Please describe the tests that you ran to verify your changes.
- Provide instructions so we can reproduce.
- Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

# Checklist:
Prepare items below using:

- [ ] updated the lock file.
- [ ] I have added tests that prove my fix is effective or that my feature works.


riordache12:
@MAfarrag - let's discuss this when you are back, I need to pick your brains on it.

MAfarrag:
@riordache12 is this ready for review? if yes please change it from draft to ready for review

PR #1162, opened 2026-07-30
fix(models): add filetype validation for ExtOldFileType enum and inte…

### LINKED ISSUE #753: `ExtOldForcing` does not have a validator for filetype
**Describe the bug**
- The `ExtOldForcing`  does not have a validator for the `filetype` parameter.
- The filetype parameter is used in a root validator to choose the right class for the file (`DiskOnlyFileModel`, `PolyFile`, `TimModel`), and wrong filetype will crash the converter to the new external forcing format.

**Expected behavior**
a field validator to check the given value for the `filetype` parameter is within the possible values.

 - Version 0.8.1

riordache12:
# Description

Added an explicit validate_filetype field validator to ExtOldForcing, plus a defensive guard in choose_file_model:
- Raises a clear ValueError if the value is not an integer (e.g. "not_a_number")
- Raises a clear ValueError if the integer is not a known filetype (e.g. 99)
- Adds a try/except in choose_file_model so that if validation of filetype fails mid-validation, the filename resolution doesn't crash with an unhandled exception

- Fixes #753
- 
Check relevant points.

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

TestValidateFileType — parametrized over all valid enum values (both enum and int forms), plus invalid integer and non-integer error cases.

# Checklist:
Prepare items below using:

- [ ] updated the lock file.
- [x] New and existing unit tests pass locally with my changes.


hydrolib/core/dflowfm/extold/models.py
```diff
@@ -352,6 +353,44 @@ def validate_quantity(cls, value) -> Any:
...
+
+        Returns:
+            ExtOldFileType: The validated filetype enum member.
+
+        Raises:
+            ValueError: If the value cannot be converted to an integer or is not a
+                valid ExtOldFileType value.
+        """
+        if isinstance(value, ExtOldFileType):
+            return value
```
MAfarrag: can you refactor this to have one return statement ?

hydrolib/core/dflowfm/extold/models.py
```diff
@@ -352,6 +353,44 @@ def validate_quantity(cls, value) -> Any:
...
+
+        valid_values = [e.value for e in ExtOldFileType]
```
MAfarrag: can you rename the `e` in the comprehension here, it is no descriptive at all

hydrolib/core/dflowfm/extold/models.py
```diff
@@ -352,6 +353,44 @@ def validate_quantity(cls, value) -> Any:
...
+        if isinstance(value, bool):
```
MAfarrag: why specifically are you checking for bool ?

hydrolib/core/dflowfm/extold/models.py
```diff
@@ -352,6 +353,44 @@ def validate_quantity(cls, value) -> Any:
...
+            )
+
+        try:
+            int_value = int(value)
+        except (ValueError, TypeError):
+            raise ValueError(
+                f"FILETYPE '{value}' is not a valid integer. Supported values: {valid_values}."
+            )
+
+        enum_value = ExtOldFileType._value2member_map_.get(int_value)
```
MAfarrag: it is not desirable to use private methods from other packages, these are not public api and can break even in minor and patch versions

MAfarrag: also why not just `int_value in ExtOldFileType` this will give you `True` or `False` , if `False` raise the error, if `True` get the value `ExtOldFileType(int_value).value`

hydrolib/core/dflowfm/extold/models.py
```diff
@@ -521,7 +560,13 @@ def choose_file_model(cls, values: Dict[str, Any]) -> Dict[str, Any]:
             raw_path = values.get(filename_var_name)
 
             if isinstance(raw_path, (Path, str)):
-                model = FILETYPE_FILEMODEL_MAPPING.get(int(file_type))
+                try:
```
MAfarrag: can you create a utility function `is_int` and add the these couple of line in it, also use it above in the `validate_filetype` validator.

MAfarrag: there are multiple return statements here, it is very hard to read

MAfarrag (changes requested):

hydrolib/core/dflowfm/extold/models.py
riordache12: I was checking for bool because `int(True)` and `int(False)` are not raising a `ValueError`

MAfarrag (approved):

PR #1178, opened 2026-08-24
feature(quantities): add lateraldischarge quantities

### LINKED ISSUE #974: Unsupported quantity `LateralDischarge`
**Describe the bug**
`LateralDischarge`, `lateraldischarge1d` and `lateraldischarge2d` are quantities that are supported in the kernel but not yet by the converter.
Refer to the user manual on Lateral Discharge quantities:
https://content.oss.deltares.nl/delft3dfm2d3d/D-Flow_FM_User_Manual.pdf
section: C.6.2.2 Lateral discharge definitions

# Related
- [UNST-8011](https://issuetracker.deltares.nl/browse/UNST-8011)

**To Reproduce**
`p:\dflowfm\users\veenstra\extforce-convert\dflowfm3d_dwaq-grevelingen-j19_6-v2b\computations\hist\jaarsom_2017\` run this model and it will throw validationerrors on the LateralDischarges.

**Expected behavior**
The LateralDischarge quantity should end up under a [Lateral] header in the new ext force file.

riordache12:
# Description
- Added the following quantities:  `LateralDischarge`, `LateralDischarge1d` and `LateralDischarge2d` according to their user manual definitions

The Lateral block has the following:
- `id`
- `name`
- `locationType` (optional) (replaces the old `type`)
- `applyTransport`: defaults to 0, with options 0 or 1
- `nodeId` (optional) - only for 1d node
- `branchId` (optional) - only for a 1d point, in combination with `chainage`
- `chainage` (optional) - only for 1d point, in combination with `branchId`
- `numCoordinates` (optional) > 1 , number of values in `xCoordinates` and `yCoordinates`
- `xCoordinates` (optional)
- `yCoordinates` (optional)
- `locationFile` (optional)
- `discharge` (optional)

Added or improved: 
- validator for `LocationType` to be part of the possibilities `1d`, `2d`, `all`
- validator for `applyTransport` to be an integer, either 0 or 1

The `LateralConverter` class:
- I've added a mapping that maps the old quantity strigns to their corresponding `locationtype` (see above)
- `convert` function, takes and `ExtOldForcing` block and returns a `Lateral` object, by:
    - first, i validate whether the quantity is supported;
    - I resolve the discharge value (whether scalar, tim or polyfile)
    - extracting location data
    - returning `Lateral` instance
- `get_discharge` function, used to handle a:
    - constant value
    - converts to a .bc file using `TimToForcingConverter`
    - locates .tim files, merges them and converts to a `ForcingModel`
- `get_location_data`: derives the lateral id and location reference

Refactored the function in `util.py` to use a class called `LocationValidator` that calls `validate()` function

- Fixes #974 

Check relevant points.

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?
- added tests for the `LateralConverter`, covering the following discharge scenarios:
   - Scalar
   - Time-series from a TIM file (TimModel → ForcingModel / .bc)
   - PolyFile location with adjacent TIM file(s) → ForcingModel
   - PolyFile location with a constant scalar value (no TIM file)
   - "realtime" discharge (verified via the Lateral model validator)
   - A fully-parsed ForcingModel as discharge ("bc file")

# Checklist:
Prepare items below using:

- [ ] updated the lock file.
- [x] New and existing unit tests pass locally with my changes.


hydrolib/core/dflowfm/ext/models.py
```diff
@@ -357,37 +383,48 @@ def validate_discharge(cls, v):
...
-        """Validates that the correct location specification is given."""
+        """Validates that the correct location specification is given.
+
+        A ``locationFile`` referencing a polygon file is accepted as a complete
+        location specification on its own (no coordinates or nodeId/branchId needed).
+        All other combinations are validated by the generic
+        :func:`validate_location_specification` helper.
+        """
+        raw_loc_file = values.get("locationfile") or values.get("locationFile")
+        if _is_non_null_location_file(raw_loc_file):
```
MAfarrag: can you refactor this to have one return statement ?

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,337 @@ class LocationValidationFieldNames(BaseModel):
...
+            or self._has_x_coordinates
+            or self._has_y_coordinates
+            or self._has_num_coordinates
         )
-        return has_node_id and not has_other
+        return self._has_node_id and not has_other
 
-    def is_valid_branch_specification() -> bool:
+    @property
+    def _is_valid_branch_specification(self) -> bool:
```
MAfarrag: why the property is using private name here ?

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,337 @@ class LocationValidationFieldNames(BaseModel):
...
+            or self._has_num_coordinates
         )
-        return has_branch_id and has_chainage and not has_other
+        return self._has_branch_id and self._has_chainage and not has_other
 
-    def is_valid_coordinates_specification() -> bool:
-        has_other = has_node_id or has_branch_id or has_chainage or has_num_coordinates
-        return has_x_coordinates and has_y_coordinates and not has_other
+    @property
+    def _is_valid_coordinates_specification(self) -> bool:
```
MAfarrag: property name here as well, why private name ?

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,337 @@ class LocationValidationFieldNames(BaseModel):
...
-        if is_valid_node_specification():
-            if config.validate_location_type:
-                validate_location_type(LocationType.oned)
-            return values
+        Returns:
+            Optional[Dict]: The validated values dict if the node specification
+                is valid, otherwise ``None``.
+        """
+        if not self._config.validate_node:
+            return None
```
MAfarrag: multiple return statements here, also return None is not desirable, need refactoring

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,337 @@ class LocationValidationFieldNames(BaseModel):
...
+        if self._is_valid_node_specification:
+            if self._config.validate_location_type:
+                self._validate_location_type_for_node_or_branch(LocationType.oned)
+            return self._values
+        error_parts.append(self._fields.node_id)
+        return None
+
+    def _try_validate_branch(self, error_parts: List[str]) -> Optional[Dict]:
```
MAfarrag: why the private names everywhere ?

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,337 @@ class LocationValidationFieldNames(BaseModel):
...
-        if config.validate_num_coordinates:
-            if is_valid_coordinates_with_num_coordinates_specification():
-                validate_coordinates_with_num_coordinates()
-                return values
+        Returns:
+            Optional[Dict]: The validated values dict if the coordinate
+                specification is valid, otherwise ``None``.
+        """
+        if not self._config.validate_coordinates:
+            return None
```
MAfarrag: return None, gives a sign that the logic needs refactoring, usually you do not need to use this return None

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,337 @@ class LocationValidationFieldNames(BaseModel):
...
+        f = self._fields
+        if self._config.validate_num_coordinates:
+            return self._try_validate_coordinates_with_num_coordinates(error_parts)
```
MAfarrag: please refactor this logic

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,337 @@ class LocationValidationFieldNames(BaseModel):
...
+        f = self._fields
+        if self._is_valid_coordinates_with_num_coordinates_specification:
+            self._validate_coordinates_with_num_coordinates()
+            if self._config.validate_location_type:
+                self._validate_location_type_for_coordinates()
+            return self._values
+        error_parts.append(
+            f"{f.x_coordinates}, {f.y_coordinates} and {f.num_coordinates}"
+        )
+        return None
```
MAfarrag: again here

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,337 @@ class LocationValidationFieldNames(BaseModel):
...
+        error_parts: List[str] = []
 
-    error = " or ".join(error_parts) + " should be provided"
-    raise ValueError(error)
+        validators: List[Callable[[List[str]], Optional[Dict]]] = [
+            self._try_validate_node,
+            self._try_validate_branch,
+            self._try_validate_coordinates,
+        ]
+        for try_validate in validators:
```
MAfarrag: this logic needs refactoring, having the raise at the bottom like that is not good, make the condition clear if () else (raise error her)

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,337 @@ class LocationValidationFieldNames(BaseModel):
...
+
+    Raises:
+        ValueError: When no valid location specification is present, when
+            ``numCoordinates`` is inconsistent, or when ``locationType`` is
+            invalid for the given location type.
+
+    Returns:
+        Dict: Validated (and possibly mutated) dictionary of input class fields.
+    """
+    return LocationValidator(values, config, fields).validate()
```
MAfarrag: why this function is needed it has one line that is a direct call to the class, remove this `validate_location_specification` and use the class directly where it is called

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1096,6 +1099,303 @@ def convert(
...
+        """Derive the discharge value from the old forcing block.
+
+        Args:
+            forcing (ExtOldForcing): The old forcing block.
+            time_unit (Optional[str]): The time unit string for time series data.
+
+        Returns:
+            Any: A constant float, a ForcingModel, or a file path representing the discharge.
+        """
+        if forcing.value is not None and not isinstance(forcing.filename, PolyFile):
```
MAfarrag: please refactor these multiple return statements ?

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1096,6 +1099,303 @@ def convert(
...
+        quantity = str(forcing.quantity).lower()
+        if quantity not in self._QUANTITY_TO_LOCATION_TYPE:
+            raise ValueError(f"Unsupported lateral quantity: {forcing.quantity}")
+        location_type = self._QUANTITY_TO_LOCATION_TYPE[quantity]
+
+        # Determine discharge and location data
+        discharge = self._get_discharge(forcing, self._get_time_unit(time_unit))
+        location_data = self._get_location_data(forcing)
+
+        data: Dict[str, Any] = {"id": location_data.pop("id")}
```
MAfarrag: why did you have to add the typing here ? if needed please use the modern typing `dict[str, Any]`

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1096,6 +1099,303 @@ def convert(
...
+            raise LateralError(
+                f"Failed to create the Lateral object for the following errors: {e}"
+            )
+
+        return new_block
+
+    def _get_time_unit(self, time_unit: Optional[str]) -> Optional[str]:
+        """Return *time_unit*, falling back to the MDU reference date when available."""
+        if time_unit is not None:
+            return time_unit
```
MAfarrag: please refactor the multiple return statements ?

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1096,6 +1099,303 @@ def convert(
...
+        Returns:
+            Optional[TimModel]: The merged `TimModel` (with ``quantities_names``
+                set to ``"discharge"`` for every column), or ``None`` when no
+                ``.tim`` files are found next to the polyline.
+        """
+        resolved = resolve_relative_to_root(polyline.filepath, self.root_dir)
+        stem = polyline.filepath.stem
+        tim_files = sorted(resolved.parent.glob(f"{stem}*.tim"))
+        if not tim_files:
+            return None
```
MAfarrag: return None is a sign that this needs refactoring

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -340,6 +362,10 @@ class Lateral(INIBasedModel):
     numcoordinates: Optional[int] = Field(None, alias="numCoordinates")
     xcoordinates: Optional[List[float]] = Field(None, alias="xCoordinates")
     ycoordinates: Optional[List[float]] = Field(None, alias="yCoordinates")
+    locationfile: Optional[
```
MAfarrag: can you update the typing `Optional` in the Lateral class to `|` also the old `List` to list , ....

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -340,6 +362,10 @@ class Lateral(INIBasedModel):
...
+        Annotated[DiskOnlyFileModel, BeforeValidator(set_default_disk_only_file_model)]
```
MAfarrag: why did you need to add this before validator ?

MAfarrag (changes requested):

hydrolib/core/dflowfm/ini/util.py
riordache12: Because I've refactored the former function to be a class. These functions are only meant to be used internally and the only function exposed is `validate()`. The rest are only being called inside this class.

hydrolib/core/dflowfm/ini/util.py
riordache12: See above

hydrolib/core/dflowfm/ext/models.py
riordache12: Well it's the same like in other classes on `locationfile` fields, like in the `Boundary` class for example. I used it to make sure that the input is validated to point to a path from where a `DiskOnlyFileModel` can be constructed. Why do you not think this is necessary?

riordache12:
See my comments where applicable. If you still don't agree with my explanations then we can sit down and have a small technical discussions.

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,310 @@ class LocationValidationFieldNames(BaseModel):
...
 
-    # `mode="before"` validators receive the raw input dict before Pydantic
-    # has resolved aliases, so a user passing the camelCase alias (e.g.
-    # `branchId`) leaves the key in its original case. The lookups below all
-    # use the lowercase field-name form, so normalize alias keys here.
-    if isinstance(values, dict):
+    def __init__(
+        self,
+        values: Dict,
+        config: Optional[LocationValidationConfiguration] = None,
```
MAfarrag: please update the old typing style

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,310 @@ class LocationValidationFieldNames(BaseModel):
...
+        fields: Optional[LocationValidationFieldNames] = None,
+    ) -> None:
+        self._config = config if config is not None else LocationValidationConfiguration()
```
MAfarrag: the entire attributes are private, in a sense the entire class is not going to be exported to users from the `__all__` so it is all private, so to create a entire class like this and make almost all the methods and attributes does not look good

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,310 @@ class LocationValidationFieldNames(BaseModel):
...
         else:
             result = len(to_list(value))
         return result
 
-    def validate_location_type(expected_location_type: LocationType) -> None:
-        location_type = values.get(fields.location_type.lower(), None)
+    def _validate_location_type_for_node_or_branch(
+        self, expected: LocationType
+    ) -> None:
+        """Validate / default ``locationType`` for node- or branch-based specs.
```
MAfarrag: using double backticks like `` something`` is a common copilot artifact, and is not needed to highlight a word please use one backtick like this `something`

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,310 @@ class LocationValidationFieldNames(BaseModel):
...
+            self._values[f.location_type.lower()] = LocationType.all
+        elif location_type not in (LocationType.oned, LocationType.twod, LocationType.all):
             raise ValueError(
-                f"{fields.x_coordinates} and {fields.y_coordinates} should have an equal amount of coordinates"
+                f"{f.location_type} has invalid value '{location_type}'. "
+                f"Possible values are: 1d, 2d, all"
             )
 
-        validate_minimum_num_coordinates(len_x_coordinates)
+    def _validate_minimum_num_coordinates(self, actual_num: int) -> None:
```
MAfarrag: again the excessive use of private method convention does not look so professional

hydrolib/core/dflowfm/ini/util.py
MAfarrag: please check again this excessive use of private convention is not professional, and almost everything in the class is private, you can just not export the class in the `__all__` from the module and the entire class will be private, but this excessive usage of the underscore is not professional, there is almost no normal method or attribute in the class  except the validate method.

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,310 @@ class LocationValidationFieldNames(BaseModel):
...
+            Dict: The (possibly mutated) values dict, with ``locationType``
+            defaulted where applicable.
 
-    error = " or ".join(error_parts) + " should be provided"
-    raise ValueError(error)
+        Raises:
+            ValueError: When no valid location specification can be found in
+                the values dict, or when a sub-validator detects an inconsistency.
+        """
+        error_parts: List[str] = []
```
MAfarrag: why did you have to declare the typing inside the method here ?

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,310 @@ class LocationValidationFieldNames(BaseModel):
...
+            self._try_validate_node,
+            self._try_validate_branch,
+            self._try_validate_coordinates,
+        ]
+        result = next(
+            (r for try_validate in validators if (r := try_validate(error_parts)) is not None),
+            None,
+        )
+        if result is None:
+            raise ValueError(" or ".join(error_parts) + " should be provided")
```
MAfarrag: for string injections we use fstring not +

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -324,25 +324,51 @@ def validate_operand(cls, v: Any):
...
-    xcoordinates: Optional[List[float]] = Field(None, alias="xCoordinates")
-    ycoordinates: Optional[List[float]] = Field(None, alias="yCoordinates")
+    locationtype: str | None = Field(None, alias="locationType")
+    nodeid: str | None = Field(None, alias="nodeId")
+    branchid: str | None = Field(None, alias="branchId")
+    chainage: float | None = Field(None, alias="chainage")
+    numcoordinates: int | None = Field(None, alias="numCoordinates")
+    xcoordinates: list[float] | None = Field(None, alias="xCoordinates")
+    ycoordinates: list[float] | None = Field(None, alias="yCoordinates")
+    locationfile: Annotated[
```
MAfarrag, line 347: indentation does not look correct

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -360,37 +386,48 @@ def validate_discharge(cls, v):
 
     @model_validator(mode="before")
     def validate_that_location_specification_is_correct(cls, values: Dict) -> Dict:
-        """Validates that the correct location specification is given."""
-        return validate_location_specification(
-            values, config=LocationValidationConfiguration(minimum_num_coordinates=1)
-        )
+        """Validates that the correct location specification is given.
+
+        A ``locationFile`` referencing a polygon file is accepted as a complete
```
MAfarrag: please change this copilot artifact (double backtick) to single backtick, docstring and mkdocs needs only one backtick to highlight a word

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,310 @@ class LocationValidationFieldNames(BaseModel):
...
+    This class groups the boolean presence-checks, individual sub-validators,
+    and the top-level orchestration that was previously spread across a single
+    function with many nested helpers.  It is used by the module-level
+    :func:`validate_location_specification` convenience wrapper.
 
     Args:
         values (Dict):
-            Dictionary of object's validated fields.
+            Dictionary of object's validated fields (mutated in-place when
+            defaulting ``locationType``).
```
MAfarrag: the backticks here

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -472,183 +472,310 @@ class LocationValidationFieldNames(BaseModel):
...
+
+        Handles both the ``numCoordinates``-present and ``numCoordinates``-absent
+        sub-cases.
+
+        Args:
+            error_parts (List[str]): Accumulator for error message fragments.
 
-        error_parts.append(f"{fields.branch_id} and {fields.chainage}")
+        Returns:
+            Optional[Dict]: The validated values dict if the coordinate
```
MAfarrag: update the docstring

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1096,6 +1099,302 @@ def convert(
...
+            tim_model,
+            time_unit,
+            units=units,
+            user_defined_names=user_defined_names,
+        )
+        forcing_model = ForcingModel(forcing=time_series_list)
+        forcing_model.filepath = location_file.with_suffix(".bc")
+        return forcing_model
+
+    def _get_location_data(self, forcing: ExtOldForcing) -> dict[str, Any]:
```
MAfarrag: this method looks like it belong to the `ExtOldForcing` class, everything inside it operates on the `ExtOldForcing`, no usage to self at all, can you move it to the  `ExtOldForcing`?

MAfarrag: before moving it to the `ExtOldForcing` can you check also if this is could be used  somewhere else ?

MAfarrag (changes requested):

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -450,183 +450,315 @@ class LocationValidationFieldNames(BaseModel):
...
+        """Validate / default `locationType` for coordinate-based specs.
 
-        if len_x_coordinates != len_y_coordinates:
+        When `xCoordinates` and `yCoordinates` are given, `locationType`
+        may be `1d`, `2d` or `all`.  Absent defaults to `all`.
+        """
+        f = self.fields
+        location_type = self.values.get(f.location_type.lower(), None)
+        if str_is_empty_or_none(location_type):
+            self.values[f.location_type.lower()] = LocationType.all
```
MAfarrag: why this hard coding of the `LocationType.all`, this assign the LocationType.all as a value even if the user did not have it. this is different than the behaviour in the function before refactoring.

why did you have to assign it.

like this any ObservationPoint that does not have a locationtype will be assigned a locationtype = all

MAfarrag: so reading this  
```ini
[ObservationPoint]
name = p1
x    = 1.0
y    = 2.0
```
as `ObservationPointModel` and saving it, will change it and assign the location type
```ini
[ObservationPoint]
name = p1
locationType = all
x    = 1.0
y    = 2.0
```
which the validator should not do, validator does not alter the values

hydrolib/core/dflowfm/extold/models.py
```diff
@@ -330,6 +330,37 @@ class ExtOldForcing(BaseModel):
...
+            isinstance(self.filename, DiskOnlyFileModel)
+            and self.filename.filepath is not None
+        ):
+            result = self.filename.filepath.stem
+        else:
+            result = str(self.quantity)
+
+        return result
+
+    def get_location_filepath(self) -> Path | None:
```
MAfarrag: why is this method? it is only called one time, why not just calling the `self.filename.filepath` directly at the place where it is called ?

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,314 @@ def convert(
...
+        """Extract location data from the old forcing block.
+
+        Args:
+            forcing (ExtOldForcing): The old forcing block.
+
+        Returns:
+            Dict[str, Any]: A dict with 'id' and either a 'locationfile' key (when
+                the source is a PolyFile) or inline coordinate fields.
+        """
+        location_filepath = (forcing.get_location_filepath() if
```
MAfarrag: can you fix the indentation ?

tests/data/input/lateral/lateral.ext
```diff
@@ -0,0 +1,67 @@
+**************************************************************************************************************
```
MAfarrag: can you remove this header ?

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,314 @@ def convert(
...
+            time_unit (Optional[str]): The time unit string for time series data.
+
+        Returns:
+            Any: A constant float, a ForcingModel, or a file path representing the discharge.
+        """
+        result = forcing.value
+
+        if forcing.value is None:
+            if isinstance(forcing.filename, TimModel):
+                result = self._get_discharge_from_tim_model(forcing, time_unit)
```
MAfarrag: `forcing.filename` will actually never be a `TimModel` it is always `PolyFile`, so this is a dead code, can you please check ?

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,314 @@ def convert(
...
+            elif isinstance(forcing.filename, PolyFile):
```
MAfarrag: this is correct method that derives the name of the tim file from the polyfile, then go and read that time file, but the upper branch `isinstance(forcing.filename, TimeModel)` this will never happen, can you please check ?, also there is no test for the this above branch, as there is no old ext file that has the field `FILENAME=<something.tim>`

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,314 @@ def convert(
...
+        location_type = self._QUANTITY_TO_LOCATION_TYPE[quantity]
+
+        discharge = self._get_discharge(forcing, self._get_time_unit(time_unit))
+        location_data = self._get_location_data(forcing)
+
+        data = {
+            "id": location_data.pop("id"),
+            "name": forcing.quantity
+        }
+        if location_type is not None:
```
MAfarrag: the variable `location_type` will never be None, as it has a value already from the line 1219 `  location_type = self._QUANTITY_TO_LOCATION_TYPE[quantity]` so this condition is not needed, can you please check if this is the intended behavior?

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,314 @@ def convert(
...
+            ValueError: If neither a TIM file nor a constant value can be found.
+        """
+        location_file = forcing.filename.filepath
+        tim_model = self._resolve_tim_file(forcing.filename, forcing.quantity)
+
+        if tim_model is not None:
+            result = self._convert_poly_tim_to_forcing_model(
+                tim_model, location_file, time_unit
+            )
+        elif forcing.value is not None:
```
MAfarrag: this value `forcing.value` will always be None,  as the condition to enter the `_get_discharge_from_poly_file` is that the `forcing.value is None`, so this is a dead code.

```python
# this is at line 1303
if forcing.value is None:
    if isinstance(forcing.filename, TimModel):
        result = self._get_discharge_from_tim_model(forcing, time_unit)
    elif isinstance(forcing.filename, PolyFile):
        result = self._get_discharge_from_poly_file(forcing, time_unit)
```

please check.

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -356,37 +381,42 @@ def validate_discharge(cls, v):
...
-    def validate_location_type(cls, v: str) -> str:
-        """
-        Method to validate whether the specified location type is correct.
-
-        Args:
-            v (str): Given value for the locationtype field.
-
-        Raises:
-            ValueError: When the value given for locationtype is unknown.
+    def validate_applytransport(cls, v: Any) -> int:
```
MAfarrag: why validating the raw data here with a before validator ? an after validator is enough

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -325,25 +323,52 @@ def validate_operand(cls, v: Any):
...
+    branchid: str | None = Field(None, alias="branchId")
+    chainage: float | None = Field(None, alias="chainage")
+    numcoordinates: int | None = Field(None, alias="numCoordinates")
+    xcoordinates: list[float] | None = Field(None, alias="xCoordinates")
+    ycoordinates: list[float] | None = Field(None, alias="yCoordinates")
+    locationfile: (
+        Annotated[DiskOnlyFileModel, BeforeValidator(set_default_disk_only_file_model)]
+        | None
+    ) = Field(None, alias="locationFile")
+    applytransport: int = Field(0, alias="applyTransport")
```
MAfarrag: can you make this field optional and change the default value to None, default value of `0` makes reading a quantity that does not have this `applytransport` field at all from a file then saving this file again will add it to the file even though it was not there before

MAfarrag (changes requested):

hydrolib/tools/extforce_convert/converters.py
riordache12: Done

tests/data/input/lateral/lateral.ext
riordache12: Done

hydrolib/core/dflowfm/extold/models.py
riordache12: Done

hydrolib/core/dflowfm/ext/models.py
riordache12: I have refactored it to be optional for now, but please double check with Julien, since in the manual it is not explicitly stated.

riordache12: Reverted this to still be 0 and non-optional, as per the Teams discussion

hydrolib/core/dflowfm/ext/models.py
riordache12: Done, it is now an `after` validator that only checks if v is not 0 or 1

hydrolib/core/dflowfm/ini/util.py
riordache12: Good catch, it was a mistake on my side, from wanting to default to `all`

hydrolib/tools/extforce_convert/converters.py
riordache12: Solved for now but please double check with Julien the preferred type on the 3 possible quantitites: `lateraldischarge1d`, `lateraldischarge2d` and `lateraldischarge`

hydrolib/tools/extforce_convert/converters.py
riordache12: Removed this branch altogether

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,276 @@ def convert(
...
+        Raises:
+            LateralError: If the Lateral object could not be created.
+        """
+        quantity = str(forcing.quantity).lower()
+
+        self.check_lateral_quantity(quantity)
+
+        location_type = self._QUANTITY_TO_LOCATION_TYPE[quantity]
+
+        discharge = self._get_discharge(forcing, self._get_time_unit(time_unit))
```
MAfarrag: why not assign this directly at line 1230 instead of creating the variable `discharge`, then use it again at line 1230 ?

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -397,6 +400,7 @@ class SourceSink(CoordinateValidator, INIBasedModel):
     All lowercased attributes match with the source-sink input as described in
     [UM Sec.C.5.2.4](https://content.oss.deltares.nl/delft3dfm1d2d/D-Flow_FM_User_Manual_1D2D.pdf#subsection.C.5.2.4).
     """
+
```
MAfarrag: can you revert these changes ? I believe these are copilot artifacts, copilot sometimes run the formatters we have in the pre-commit hooks file and then you we get these un intended changes

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -780,9 +784,7 @@ class Comments(SpatialForcingBase.Comments):
             "Name of file containing the data for this spatial quantity.",
             alias="dataFile",
         )
-        datafiletype: str | None = Field(
-            "Type of dataFile.", alias="dataFileType"
-        )
+        datafiletype: str | None = Field("Type of dataFile.", alias="dataFileType")
```
MAfarrag: can you revert these changes ?

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -820,7 +822,8 @@ class Comments(SpatialForcingBase.Comments):
             alias="dataValue",
         )
         frictiontype: str | None = Field(
-            "Only for quantity=frictionCoefficient. The friction type.", alias="frictionType"
+            "Only for quantity=frictionCoefficient. The friction type.",
```
MAfarrag: can you revert these changes ?

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -844,7 +847,9 @@ class Comments(SpatialForcingBase.Comments):
     )
     datafiletype: DataFileType | None = Field(None, alias="dataFileType")
     datavariablename: str | None = Field(None, alias="dataVariableName")
-    targetmaskfile: PolyFile | DiskOnlyFileModel | None = Field(None, alias="targetMaskFile")
+    targetmaskfile: PolyFile | DiskOnlyFileModel | None = Field(
```
MAfarrag: can you revert these changes ?

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -857,7 +862,9 @@ class Comments(SpatialForcingBase.Comments):
     averagingtype: AveragingType | None = Field(None, alias="averagingType")
     averagingrelsize: NonNegativeFloat | None = Field(None, alias="averagingRelSize")
     averagingnummin: PositiveInt | None = Field(None, alias="averagingNumMin")
-    averagingpercentile: NonNegativeFloat | None = Field(None, alias="averagingPercentile")
+    averagingpercentile: NonNegativeFloat | None = Field(
```
MAfarrag: can you revert these changes ?

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -912,9 +919,13 @@ def _validate_datafile_path(
     ) -> None:
         """Validate the ``dataFile`` usage path and emit deprecation warning when needed."""
         if not has_datafile:
-            raise ValueError("'dataFile' is required when 'dataValue' is not specified.")
+            raise ValueError(
```
MAfarrag: can you revert these changes ?

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -912,9 +919,13 @@ def _validate_datafile_path(
...
+                "'dataFile' is required when 'dataValue' is not specified."
+            )
         if not has_datafiletype:
-            raise ValueError("'dataFileType' is required when 'dataValue' is not specified.")
+            raise ValueError(
```
MAfarrag: can you revert these changes ?

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -991,7 +1002,9 @@ def validate_datavalue_or_datafile(cls, values: Dict) -> Dict:
 
         datavalue = values.get("datavalue")
         has_datafile = (values.get("datafile") or values.get("dataFile")) is not None
-        has_datafiletype = (values.get("datafiletype") or values.get("dataFileType")) is not None
+        has_datafiletype = (
```
MAfarrag: can you revert these changes ?

hydrolib/core/dflowfm/extold/models.py
```diff
@@ -330,6 +330,34 @@ class ExtOldForcing(BaseModel):
...
+        2. TimModel stem.
+        3. Any file model exposing a non-null `filepath` stem.
+        4. Fallback to the quantity string.
+        """
+        if isinstance(self.filename, PolyFile):
+            result = self.filename.filepath.stem
+            if self.filename.objects:
+                first_obj = self.filename.objects[0]
+                if first_obj.metadata and first_obj.metadata.name:
+                    result = first_obj.metadata.name
```
MAfarrag: what is this logic ? from line 344 - 347 ?

hydrolib/core/dflowfm/extold/models.py
```diff
@@ -330,6 +330,34 @@ class ExtOldForcing(BaseModel):
...
+            result = self.filename.filepath.stem
+        elif (
+            isinstance(self.filename, DiskOnlyFileModel)
+            and self.filename.filepath is not None
+        ):
+            result = self.filename.filepath.stem
+        else:
+            result = str(self.quantity)
+
+        return result
```
MAfarrag: the logic in this function should be simplified, all the cases are `pathlib.Path` so you are getting `filename.filepath.stem` almost from al of them except the polyfile at the top and the else at the end which i do not know if they were in the original logic before you moved it here as a method, please check those two cases, and simplify the rest.

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -449,184 +452,343 @@ class LocationValidationFieldNames(BaseModel):
...
+
+        Only `1d` is accepted; `2d` and `all` are rejected because they
+        require coordinate fields to be meaningful.
+        """
+        f = self.fields
+        location_type = self.values.get(f.location_type.lower(), None)
         if str_is_empty_or_none(location_type):
-            values[fields.location_type.lower()] = expected_location_type
-        elif location_type != expected_location_type:
+            self.values[f.location_type.lower()] = expected
```
MAfarrag, line 561: the locationtype is still being assigned, and this time not with `LocationType.all` but with `LocationType.oned`, so the method  `validate_location_type_for_node_or_branch` is called in two different other locations and in both calls a value of `LocationType.oned` is passed.

tests/data/input/lateral/lateral.ext
MAfarrag: remove the whole header not just the top line, leave only the quantity blocks

hydrolib/tools/extforce_convert/main_converter.py
```diff
@@ -414,6 +415,8 @@ def _convert_forcing(
             new_quantity_block = converter_class.convert(forcing)
         elif isinstance(converter_class, MassBalanceAreaConverter):
             new_quantity_block = converter_class.convert(forcing)
+        elif isinstance(converter_class, LateralConverter):
+            new_quantity_block = converter_class.convert(forcing)
```
MAfarrag, line 419: here the converter_class.convert takes `time_unit` input parameter, but you never actually pass it here

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,276 @@ def convert(
...
+
+        Raises:
+            LateralError: If the quantity is not supported by this converter.
+        """
+        if str(quantity).lower() not in LateralConverter._QUANTITY_TO_LOCATION_TYPE:
+            raise LateralError(f"Unsupported lateral quantity: {quantity}")
+
+
+    def convert(
+        self, forcing: ExtOldForcing, time_unit: str | None = None
```
MAfarrag: this `time_unit` input parameter is never passed in the call at line 419, so it will always be None

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,276 @@ def convert(
...
+            raise LateralError(
+                f"Failed to create the Lateral object for the following errors: {e}"
+            )
+
+        return new_block
+
+    def _get_time_unit(self, time_unit: str | None) -> str | None:
+        """Return `time_unit`, falling back to the MDU reference date when available."""
+        result = time_unit
+        if result is None and self._mdu_parser is not None:
```
MAfarrag: this `time_unit` is always None, and the `_mdu_parser` will never be None, so the condition  `if result is None and self._mdu_parser is not None:` is always True.

we can never use the converter without the mdu_parser, it is the entry point , so please recheck this area

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,276 @@ def convert(
...
+            temperature_salinity_data = self._mdu_parser.temperature_salinity_data
+            if temperature_salinity_data is not None:
```
MAfarrag: can you also check this `temperature_salinity_data` it will never be None  also

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,276 @@ def convert(
...
+        Args:
+            forcing (ExtOldForcing): The old forcing block.
+            time_unit (Optional[str]): The time unit string for time series data.
+
+        Returns:
+            Any: A constant float, a ForcingModel, or a file path representing the discharge.
+        """
+        result = forcing.value
+
+        if forcing.value is None and isinstance(forcing.filename, PolyFile):
```
MAfarrag: for laterals the `focring.filename` is always `PolyFile` so this will always be True, you do not have to always do this check, can you also check in the whole branch, it is repeated.

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,276 @@ def convert(
...
+            result = self._get_discharge_from_poly_file(forcing, time_unit)
```
MAfarrag: can you merge this `_get_discharge_from_poly_file` method into the `_get_discharge` method? before there were two branches inside the `_get_discharge` now only one the polyfile condition, so no need for the extra method

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,276 @@ def convert(
...
+
+        Args:
+            forcing (ExtOldForcing): The old forcing block.
+
+        Returns:
+            Dict[str, Any]: A dict with 'id' and either a 'locationfile' key (when
+                the source is a PolyFile) or inline coordinate fields.
+        """
+        location_filepath = forcing.filename.filepath if isinstance(
+            forcing.filename, PolyFile) else None
```
MAfarrag: again here about checking if the filename is a polyfile .

hydrolib/core/dflowfm/extold/models.py
MAfarrag: sorry so this is used only in for the laterals where the filename is only polyfile right? so the original block of code before moving it here did not need any of the other branches (TimModel/DiskOnlyFileModel), can you double check this and if it is only used in Laterals and all of these other conditions are actually not needed, please clean them and move this further to the `PolyFile` class since now it is actually about the polyfile, also rename it to `get_id`

hydrolib/core/dflowfm/extold/models.py
MAfarrag: ignore this if the comment above resulted in the move to `PolyFile`

hydrolib/core/dflowfm/extold/models.py
MAfarrag: also check what is better `PolyFile` or `PolyObject` to have this new logic

hydrolib/tools/extforce_convert/converters.py
```diff
@@ -1130,6 +1131,276 @@ def convert(
...
+        discharge = self._get_discharge(forcing, self._get_time_unit(time_unit))
+        location_data = self._get_location_data(forcing)
+
+        data = {
+            "id": location_data.pop("id"),
+            "name": forcing.quantity,
+            "locationtype": location_type,
+        }
+        data.update(location_data)
+        data["discharge"] = discharge
```
MAfarrag: can you simplify the above logic by 
```python
{
    **location_data,
    "name": forcing.quantity,
    "locationtype": location_type,
    "discharge": discharge
}
```
check first it is giving the same behavior

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -356,38 +359,38 @@ def validate_discharge(cls, v):
...
-        """
-        possible_values = ["1d", "2d", "all"]
-        if v.lower() not in possible_values:
-            raise ValueError(
-                "Value given ({}) not accepted, should be one of: {}".format(
-                    v, ", ".join(possible_values)
-                )
-            )
+    def validate_applytransport(cls, v: Any) -> int:
+        if v not in (0, 1):
```
MAfarrag: can you also account for the case of None, now it is optional, so None is a valid value

hydrolib/core/dflowfm/ext/models.py
```diff
@@ -356,38 +359,38 @@ def validate_discharge(cls, v):
...
     def validate_that_location_specification_is_correct(cls, values: Dict) -> Dict:
-        """Validates that the correct location specification is given."""
-        return validate_location_specification(
-            values, config=LocationValidationConfiguration(minimum_num_coordinates=1)
+        """Validates that the correct location specification is given.
+
+        A `locationFile` referencing a polygon file is accepted as a complete
+        location specification on its own (no coordinates or nodeId/branchId needed).
+        All other combinations are validated by the generic
+        :func:`validate_location_specification` helper.
```
MAfarrag: can you update this mention of the old `validate_location_specification` to the `LocationValidationConfiguration`

hydrolib/core/dflowfm/ini/util.py
```diff
@@ -449,184 +452,343 @@ class LocationValidationFieldNames(BaseModel):
...
-            - branchId with chainage
-            - xCoordinates with yCoordinates
-            - xCoordinates with yCoordinates and numCoordinates.
-        ValueError: When numCoordinates does not meet the requirement minimum amount or does not match the amount of xCoordinates or yCoordinates.
-        ValueError: When locationType should be 1d but other was specified.
-
-    Returns:
-        Dict: Validated dictionary of input class fields.
+            Field-name overrides.  Defaults to
+            :class:`LocationValidationFieldNames` with standard D-FLOW FM names.
```
MAfarrag: can you check this docstring it still reference the old `validate_location_specification` function

MAfarrag (changes requested):

hydrolib/tools/extforce_convert/converters.py
riordache12: Done

hydrolib/core/dflowfm/ext/models.py
riordache12: Done

hydrolib/core/dflowfm/ext/models.py
riordache12: Done

hydrolib/core/dflowfm/ext/models.py
riordache12: Done

hydrolib/core/dflowfm/ext/models.py
riordache12: Done

hydrolib/core/dflowfm/ext/models.py
riordache12: Done

hydrolib/core/dflowfm/ext/models.py
riordache12: Done

hydrolib/core/dflowfm/ext/models.py
riordache12: Done

hydrolib/core/dflowfm/ext/models.py
riordache12: Done

hydrolib/core/dflowfm/extold/models.py
riordache12: Good catch, so indeed `Lateral`was the only place that was using this function. I've removed it from `extold` and added a simplified version of it inside `Polyfile` as the conversion path expects polyline-based location data. The extra `TimModel/DiskOnlyFileModel` were not needed anymore. Also renamed it to `get_id`

hydrolib/core/dflowfm/extold/models.py
riordache12: Ignored since I have moved it to Polyfile

hydrolib/tools/extforce_convert/converters.py
MAfarrag: the `lateraldischarge` is all but still we do not explicitly set it in the file

hydrolib/core/dflowfm/ini/util.py
riordache12: The logic has been maintained from the function pre-refactoring so we've decided to not touch it now.

hydrolib/tools/extforce_convert/converters.py
riordache12: It does behave identically

hydrolib/core/dflowfm/ext/models.py
riordache12: Done

hydrolib/tools/extforce_convert/converters.py
riordache12: Done

hydrolib/core/dflowfm/ini/util.py
riordache12: Done

tests/tools/test_converter_lateral.py
```diff
@@ -0,0 +1,260 @@
...
+		workspace = tmp_path / "lateral"
+		workspace.mkdir()
+
+		for name in [
+			"lateral.ext",
+			"afstroming.pol",
+			"afstroming.tim",
+			"rainfall_minus_evaporation.pol",
+			"rainfall_minus_evaporation.tim",
+		]:
```
MAfarrag: can you move these couple of lines that copy the data to the temp dir to a separate fixture ?

MAfarrag (approved):

PR #1195, opened 2026-09-08
Fix/output section does not exist waqbalancearea

### LINKED ISSUE #1196: extforce_convert: mass balance area conversion crashes when the MDU has no [output] section
**Describe the bug**

Converting an old-style external forcing model that contains `waqmassbalancearea<name>` quantities fails
with a `TypeError` when the MDU file has no `[output]` section. After moving the areas into the new
`<*_mba.ini>` file, the converter writes `mbaFile`/`mbaInterval` into the MDU `[output]` section; if that
section does not exist, `MDUParser.update_file_entry` computes the insertion line as
`section.last_key_value_line_index + 1`, where `last_key_value_line_index` is `None` for a missing section,
raising `TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'`.

The `[output]` section is optional in D-Flow FM (all its keywords have defaults), and the old mass-balance
configuration lives entirely outside it (the areas are the `waqmassbalancearea` quantities in the old ext,
and the interval is `[processes] DtMassBalance`/`DtProcesses`). So a legitimate legacy model can have mass
balance areas and no `[output]` section, and the converter crashes on it.

**To Reproduce**

End-to-end: run `extforce_convert --mdufile model.mdu` (or `ExternalForcingConverter.from_mdu(...)`) on an
old-style model whose `.ext` has a `waqmassbalancearea<name>` quantity, with `DtMassBalance` in
`[processes]` and **no `[output]` section** in the MDU. Conversion aborts with
`TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'`.

Minimal reproduction of the root cause (the line the converter hits when writing `mbaFile` after
conversion) — writing any keyword to an absent section fails:

```python
import tempfile
from pathlib import Path
from hydrolib.tools.extforce_convert.mdu_parser import MDUParser

d = Path(tempfile.mkdtemp())
mdu = d / "model.mdu"
mdu.write_text(
    "[geometry]\nNetFile = net.nc\n\n"
    "[physics]\nSalinity = 0\nTemperature = 0\n\n"
    "[time]\nRefDate = 20010101\nTunit = S\n\n"
    "[external forcing]\nExtForceFile = old.ext\n"
)  # note: no [output] section

parser = MDUParser(mdu)
parser.update_mba_file("new_mba.ini")   # <-- TypeError on v1.1.0: unsupported operand ... 'NoneType' and 'int'
```

**Expected behavior**

The converter should create the `[output]` section when it is absent and write `mbaFile`/`mbaInterval`
into it, instead of crashing.

**Version info**
 - OS: any
 - Version: 1.1.0

**Additional context**

Root cause is in `MDUParser.update_file_entry`, so any `update_*_file` writing to a section that can be
absent is exposed; mass balance area conversion is where it surfaces because `[output]` is the target and
is optional. Introduced with the mass balance area conversion feature (#1173 / #1191).

MAfarrag:
# Description

- Please include a summary of the change and which issue is fixed.
- Please also include relevant motivation and context.
- List any dependencies that are required for this change.

# Issues
- Fixes #1196 
## Type of change

Check relevant points.

# How Has This Been Tested?

- Please describe the tests that you ran to verify your changes.
- Provide instructions so we can reproduce.
- Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

# Checklist:
Prepare items below using:

- [ ] updated the lock file.
- [ ] I have added tests that prove my fix is effective or that my feature works.


hydrolib/tools/extforce_convert/mdu_parser.py
```diff
@@ -779,8 +779,13 @@ def update_file_entry(
...
             section = self.get_section(section_name)
-            # put the inifield file at the end of the geometry section
-            line_number = section.last_key_value_line_index + 1
+            if section.start is None:
+                # the target section does not exist yet (e.g. an MDU without an [output]
+                # section); create it and add the entry right after its header.
+                line_number = self.add_section(section_name) + 1
+            else:
+                # add the entry at the end of the existing section
+                line_number = section.last_key_value_line_index + 1
```
riordache12: I find the logic a bit duplicated here, I will try my best to explain why:
-  `get_section` looks for a section and returns it 
- then if it does not find it, `add_section` is called which calls `has_section` which calls again `get_section`

So `get_section` gets called twice to check if a new section should be added.

I would just add a new function called `ensure_section` and inside of it add this logic, such as either the section is found or a new one is create. Either way, a section is always returned. No `ValueError` gets returned anymore.

tests/tools/mdu_parser/test_mdu_parser_methods.py
```diff
@@ -336,22 +336,35 @@ def test_get_structure_file_when_mdu_has_no_entry_returns_none(self, tmp_path):
         assert result is None
 
 
+def _make_mdu_parser(content: list) -> MagicMock:
+    """Build a MagicMock ``MDUParser`` with the pure text-manipulation methods bound.
```
riordache12: double backticks

riordache12 (changes requested):
Hi @MAfarrag , I've added a minor comment and a more important one. Don't hesitate to call me if you believe I am missing something or you want to discuss it.

tests/tools/mdu_parser/test_mdu_parser_methods.py
MAfarrag: fixed

hydrolib/tools/extforce_convert/mdu_parser.py
MAfarrag: you were right the method needed refactoring, so i split it into exactly the two branches of the if condition, so the first update an existing field in the file, and the second is adding a new field that does not exist yet. 
now the big method `update_file_entry` is branched inside into two `_add_new_entry`, `_update_existing_entry` methods.

initially i had the a method called `has_section` that return a bool whither the section exist or not but now the `get_section` method superseded it (similar to the `ensure_section`).

riordache12 (approved):

